PR #16, opened 2014-12-05
Improve QuadraticStepSize

vladimir-ch:
* Use previous step size in case interpolation is not done
* Use floats.EqualWithinRel rather than own comparison
* Use interpolation only if there is a chance of interpolating the
  function well by a quadratic
* Remove MinStepSize, it is not necessary that users manipulate it

### DIFF
--- a/stepsizers.go
+++ b/stepsizers.go
@@ -10,7 +10,10 @@ import (
 	"github.com/gonum/floats"
 )
 
-const minStepSize = 1e-8
+const (
+	minimumQuadraticStepSize  = 1e-8
+	defaultQuadraticThreshold = 1e-12
+)
 
 // ConstantStepSize is a StepSizer that returns the same step size for
 // every iteration.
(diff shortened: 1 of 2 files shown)


stepsizers.go
```diff
@@ -27,63 +33,85 @@ func (c ConstantStepSize) StepSize(l Location, dir []float64) float64 {
...
-		// stepsize will be positive.
-		stepsize = 2 * (l.F - q.fPrev) / floats.Dot(l.Gradient, dir)
-		// Trim the step size to lie in [MinStepSize, 1]
-		stepsize = math.Max(q.MinStepSize, math.Min(1.01*stepsize, 1))
+func (q *QuadraticStepSize) StepSize(l Location, dir []float64) (stepSize float64) {
+	stepSizePrev := floats.Distance(l.X, q.xPrev, 2) / q.dirPrevNorm
+	projGrad := floats.Dot(l.Gradient, dir)
+
+	stepSize = 2 * stepSizePrev
+	if !floats.EqualWithinRel(q.fPrev, l.F, q.Threshold) {
```
btracey: The test would be something like "If the function values are close, and the absolute value of the gradient is small". If the function values are the same, but the gradient is large and of opposite sign, then it still makes sense to use a quadratic interpolation.

btracey:
This seems to be in good shape except for the one comment. We can look at parameter tuning when we've integrated in the other functions and see if there are any pathological cases.

stepsizers.go
vladimir-ch: I am still not convinced. Testing for "the absolute value of the gradient is small" means introducing a new constant (unless we reuse the one from bisection). Testing for "the gradients are opposite" means deciding upon an angle condition between the previous and current gradient and introducing another constant (projected gradients are always negative). All this for an extremely rare situation which this (or any other) step sizer cannot handle well anyway, because the descent directions in this case seem to be unrelated. Moreover, the optimizer has already advanced to the current location in l.X, so if the minimum (along the previous search direction) lies between the previous and current X, it is unfortunate, but irrelevant for the current search direction.

I like this step sizer less and less. CG seems to work slightly better with the step sizer waiting in the other PR, which leaves the steepest descent as the only possible user of this step sizer. However, in steepest descent the consecutive descent directions are not related at all to each other, so theoretically using this step sizer makes no sense in that case, it is just a heuristic.

btracey: Yea,  I see your point. I'm thinking too much in the mindset of a linesearch, and not enough as a step sizer. The penalty for failure here isn't that large,  so it's probably fine as is. Given all the work you've put in,  we may as well include it, given that it does work.

vladimir-ch:
Merged from the command line after a small experiment with rebasing and squashing.

PR #22, opened 2014-12-16
Add CG

vladimir-ch:
A proposal for a possible implementantion of the non-linear conjugate gradient method.

Points to consider:
* CG variants are quite verbose at the moment. We could consider renaming FletcherReeves to FR, PolakRibierePolyak to PRPPlus, etc.
* More variants can be added easily
* After a very limited testing, surprisingly, Hestenes-Stiefel seems to perform better than Polak-Ribiere.
* Restarting after n iterations without a restart and after an angle condition fails is performed always. A user setting to turn these restarting strategies on/off could be added.

#16 should be reviewed and merged first.

### DIFF
--- a/cg.go
+++ b/cg.go
@@ -1,0 +1,300 @@
+package optimize
+
+import (
+	"math"
+
+	"github.com/gonum/floats"
+)
+
+const (
+	iterationRestartFactor = 6
+	angleRestartThreshold  = -0.9
+)
+
+// CGVariant calculates the scaling parameter, β, used for updating the
+// conjugate direction in the nonlinear conjugate gradient (CG) method.
+type CGVariant interface {
+	// Init is called at the first iteration and provides a way to initialize
+	// any internal state.
+	Init(loc Location)
+	// Beta returns the value of the scaling parameter that is computed
+	// according to the particular variant of the CG method.
+	Beta(grad, gradPrev, dirPrev []float64) float64
+}
+
+// CG implements the nonlinear conjugate gradient method for solving nonlinear
+// unconstrained optimization problems. It is a line search method that
+// generates the search directions d_k according to the formula
+//  d_{k+1} = -∇f_{k+1} + β_k*d_k,   d_0 = -∇f_0.
+// Variants of the conjugate gradient method differ in the choice of the
+// parameter β_k. The conjugate gradient method usually requires fewer function
+// evaluations than the gradient descent method and no matrix storage, but
+// L-BFGS is usually more efficient.
+//
+// CG implements a restart strategy that takes the steepest descent direction
+// (i.e., d_{k+1} = -∇f_{k+1}) whenever any of the following conditions holds:
+//
+//  - A certain number of iterations has elapsed without a restart. This number
+//    is controllable via IterationRestartFactor and if equal to 0, it is set to
+//    a reasonable default based on the problem dimension.
+//  - The angle between the gradients at two consecutive iterations ∇f_k and
+//    ∇f_{k+1} is too large.
+//  - β_k given by CGVariant.Beta() is equal to zero.
+//
+// The line search for CG must yield step sizes that satisfy the strong Wolfe
+// conditions at every iteration, otherwise the generated search direction
+// might fail to be a descent direction. The line search should be more
+// stringent compared with those for Newton-like methods, which can be achieved
+// by setting the gradient constant in the strong Wolfe conditions to a small
+// value.
+//
+// See also William Hager, Hongchao Zhang, A survey of nonlinear conjugate
+// gradient methods. Pacific Journal of Optimization, 2 (2006), pp. 35-58, and
+// references therein.
+type CG struct {
+	// LinesearchMethod must satisfy the strong Wolfe conditions at every
+	// iteration. If LinesearchMethod == nil, an appropriate default is chosen.
+	LinesearchMethod LinesearchMethod
+	// Variant implements the particular CG formula for computing β_k.
+	// If Variant is nil, an appropriate default is chosen.
+	Variant CGVariant
+	// InitialStep estimates the initial line search step size, because the CG
+	// method does not generate well-scaled search directions.
+	// If InitialStep is nil, an appropriate default is chosen.
+	InitialStep StepSizer
+
+	// IterationRestartFactor determines the frequency of restarts based on the
+	// problem dimension. The negative gradient direction is taken whenever
+	// ceil(IterationRestartFactor*(problem dimension)) iterations have elapsed
+	// without a restart. For medium and large-scale problems
+	// IterationRestartFactor should be set to 1, low-dimensional problems a
+	// larger value should be chosen. Note that if the ceil function returns 1,
+	// CG will be identical to gradient descent.
+	// If IterationRestartFactor is 0, it will be set to 6.
+	// CG will panic if IterationRestartFactor is negative.
+	IterationRestartFactor float64
+	// AngleRestartThreshold sets the threshold angle for restart. The method
+	// is restarted if the cosine of the angle between two consecutive
+	// gradients is smaller than or equal to AngleRestartThreshold, that is, if
+	//  ∇f_k·∇f_{k+1} / (|∇f_k| |∇f_{k+1}|) <= AngleRestartThreshold.
+	// A value of AngleRestartThreshold closer to -1 (successive gradients in
+	// exact opposite directions) will tend to reduce the number of restarts.
+	// If AngleRestartThreshold is 0, it will be set to -0.9.
+	// CG will panic if AngleRestartThreshold is not in the interval [-1, 0].
+	AngleRestartThreshold float64
+
+	linesearch *Linesearch
+
+	restartAfter    int
+	iterFromRestart int
+
+	dirPrev      []float64
+	gradPrev     []float64
+	gradPrevNorm float64
+}
+
+func (cg *CG) Init(loc Location, funcInfo *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
+	if cg.IterationRestartFactor < 0 {
+		panic("cg: IterationRestartFactor is negative")
+	}
+	if cg.AngleRestartThreshold < -1 || cg.AngleRestartThreshold > 0 {
+		panic("cg: AngleRestartThreshold not in [-1, 0]")
+	}
+
+	if cg.LinesearchMethod == nil {
+		cg.LinesearchMethod = &Bisection{GradConst: 0.1}
+	}
+	if cg.Variant == nil {
+		cg.Variant = &HestenesStiefel{}
+	}
+	if cg.InitialStep == nil {
+		cg.InitialStep = &FirstOrderStepSize{}
+	}
+
+	if cg.IterationRestartFactor == 0 {
+		cg.IterationRestartFactor = iterationRestartFactor
+	}
+	if cg.AngleRestartThreshold == 0 {
+		cg.AngleRestartThreshold = angleRestartThreshold
+	}
+
+	if cg.linesearch == nil {
+		cg.linesearch = &Linesearch{}
+	}
+	cg.linesearch.Method = cg.LinesearchMethod
+	cg.linesearch.NextDirectioner = cg
+
+	return cg.linesearch.Init(loc, funcInfo, xNext)
+}
+
+func (cg *CG) Iterate(loc Location, xNext []float64) (EvaluationType, IterationType, error) {
+	return cg.linesearch.Iterate(loc, xNext)
+}
+
+func (cg *CG) InitDirection(loc Location, direction []float64) (stepSize float64) {
+	dim := len(loc.X)
+
+	cg.restartAfter = int(math.Ceil(cg.IterationRestartFactor * float64(dim)))
+	cg.iterFromRestart = 0
+
+	// The initial direction is always the negative gradient.
+	copy(direction, loc.Gradient)
+	floats.Scale(-1, direction)
+
+	cg.dirPrev = resize(cg.dirPrev, dim)
+	copy(cg.dirPrev, direction)
+	cg.gradPrev = resize(cg.gradPrev, dim)
+	copy(cg.gradPrev, loc.Gradient)
+	cg.gradPrevNorm = floats.Norm(loc.Gradient, 2)
+
+	cg.Variant.Init(loc)
+	return cg.InitialStep.Init(loc, direction)
+}
+
+func (cg *CG) NextDirection(loc Location, direction []float64) (stepSize float64) {
+	copy(direction, loc.Gradient)
+	floats.Scale(-1, direction)
+
+	cg.iterFromRestart++
+	var restart bool
+	// Check if the method should be restarted because too many iterations have
+	// been taken without a restart.
+	if cg.iterFromRestart == cg.restartAfter {
+		restart = true
+	}
+	// Check if the method should be restarted because the angle between
+	// the last two gradients is too large.
+	gDot := floats.Dot(loc.Gradient, cg.gradPrev)
+	gNorm := floats.Norm(loc.Gradient, 2)
+	if gDot <= cg.AngleRestartThreshold*gNorm*cg.gradPrevNorm {
+		restart = true
+	}
+
+	// Compute the scaling factor β_k according to the given CG variant. Do
+	// this even when restarting, because cg.Variant may be keeping an inner
+	// state that needs to be updated at every iteration.
+	beta := cg.Variant.Beta(loc.Gradient, cg.gradPrev, cg.dirPrev)
+	if beta == 0 {
+		// β_k == 0 means that the steepest descent direction will be taken, so
+		// indicate that the method is in fact being restarted.
+		restart = true
+	}
+	if !restart {
+		// The method is not being restarted, so update the descent direction.
+		floats.AddScaled(direction, beta, cg.dirPrev)
+	} else {
+		// Otherwise reset to 0 the counter of iterations taken since the last
+		// restart.
+		cg.iterFromRestart = 0
+	}
+
+	// Get the initial line search step size from the StepSizer, even if the
+	// method was restarted, because StepSizers need to see every iteration.
+	stepSize = cg.InitialStep.StepSize(loc, direction)
+	if restart {
+		// The method was restarted and since the steepest descent direction is
+		// not related to the previous direction, discard the estimated step
+		// size from cg.InitialStep and use step size of 1 instead.
+		stepSize = 1
+	}
+
+	copy(cg.gradPrev, loc.Gradient)
+	copy(cg.dirPrev, direction)
+	cg.gradPrevNorm = gNorm
+	return stepSize
+}
+
+// FletcherReeves implements the Fletcher-Reeves variant of the CG method that
+// computes the scaling parameter β_k according to the formula
+//  β_k = |∇f_{k+1}|^2 / |∇f_k|^2.
+type FletcherReeves struct {
+	prevNorm float64
+}
+
+func (fr *FletcherReeves) Init(loc Location) {
+	fr.prevNorm = floats.Norm(loc.Gradient, 2)
+}
+
+func (fr *FletcherReeves) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	norm := floats.Norm(grad, 2)
+	beta = (norm / fr.prevNorm) * (norm / fr.prevNorm)
+	fr.prevNorm = norm
+	return beta
+}
+
+// PolakRibierePolyak implements the Polak-Ribiere-Polyak variant of the CG
+// method that computes the scaling parameter β_k according to the formula
+//  β_k = max(0, ∇f_{k+1}·y_k / |∇f_k|^2),
+// where y_k = ∇f_{k+1} - ∇f_k.
+type PolakRibierePolyak struct {
+	prevNorm float64
+}
+
+func (pr *PolakRibierePolyak) Init(loc Location) {
+	pr.prevNorm = floats.Norm(loc.Gradient, 2)
+}
+
+func (pr *PolakRibierePolyak) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	norm := floats.Norm(grad, 2)
+	dot := floats.Dot(grad, gradPrev)
+	beta = (norm*norm - dot) / (pr.prevNorm * pr.prevNorm)
+	pr.prevNorm = norm
+	return math.Max(0, beta)
+}
+
+// HestenesStiefel implements the Hestenes-Stiefel variant of the CG method
+// that computes the scaling parameter β_k according to the formula
+//  β_k = max(0, ∇f_{k+1}·y_k / d_k·y_k),
+// where y_k = ∇f_{k+1} - ∇f_k.
+type HestenesStiefel struct {
+	y []float64
+}
+
+func (hs *HestenesStiefel) Init(loc Location) {
+	hs.y = resize(hs.y, len(loc.Gradient))
+}
+
+func (hs *HestenesStiefel) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	floats.SubTo(hs.y, grad, gradPrev)
+	beta = floats.Dot(grad, hs.y) / floats.Dot(dirPrev, hs.y)
+	return math.Max(0, beta)
+}
+
+// DaiYuan implements the Dai-Yuan variant of the CG method that computes the
+// scaling parameter β_k according to the formula
+//  β_k = |∇f_{k+1}|^2 / d_k·y_k,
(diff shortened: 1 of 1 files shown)


cg.go
```diff
@@ -0,0 +1,162 @@
+package optimize
+
```
btracey: All these exported structs need comments. I'm not familiar with CGVariant, so I personally would appreciate some explanation. Up to you what the appropriate level is, as we obviously aren't going to write a book in the comments.

cg.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/gonum/floats"
+)
+
+const (
+	iterationRestartFactor = 6
+	angleRestartTreshold   = 0.9
+)
+
+type CGVariant interface {
```
btracey: I don't particularly like the word Variant. I understand what you mean, but it's a pretty vague term. Wikipedia calls it a "Preconditioner" which is much more descriptive. If you don't think that's misleading, I'd rather see CGPreconditioner (or CGPrecon) than CGVariant.

cg.go
```diff
@@ -0,0 +1,162 @@
...
+
+	cg.Variant.Init(loc)
+	return cg.InitialStep.Init(loc, direction)
+}
+
+func (cg *CG) NextDirection(loc Location, direction []float64) (stepSize float64) {
+	copy(direction, loc.Gradient)
+	floats.Scale(-1, direction)
+
+	var β float64
```
btracey: @kortschak what is our policy on Unicode? I would probably just write beta.

cg.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+const (
+	iterationRestartFactor = 6
+	angleRestartTreshold   = 0.9
+)
+
+type CGVariant interface {
+	Init(loc Location)
+	Beta(gradient, gradPrev, dirPrev []float64) float64
```
btracey: Looking at the CG code, there should definitely be a comment on Beta specifying that zero is a special value.

cg.go
```diff
@@ -0,0 +1,162 @@
...
+	floats.Scale(-1, direction)
+
+	var β float64
+	cg.iterFromRestart++
+	restart := cg.iterFromRestart%cg.restartAfter == 0
+	if !restart {
+		β = cg.Variant.Beta(loc.Gradient, cg.gradPrev, cg.dirPrev)
+	}
+	if β != 0 {
+		floats.AddScaled(direction, β, cg.dirPrev)
```
btracey: Could you add a quick high-level comment here.

// If beta == 0, it means that it is time to restart (because of iteration number or the CGVariant). When we restart, ...

btracey: Sorry, the beginning to my comment was totally wrong. Instead it should be something like 
// We have a scaling factor from the variant. Update the direction from the CG rule
(and if it's an easy rule in math notation, say it).

cg.go
```diff
@@ -0,0 +1,162 @@
...
+		gPrevNorm := floats.Norm(cg.gradPrev, 2)
+		gDotGPrev := floats.Dot(loc.Gradient, cg.gradPrev)
+
+		if gDotGPrev <= -cg.AngleRestartTreshold*gNorm*gPrevNorm {
+			copy(direction, loc.Gradient)
+			floats.Scale(-1, direction)
+			cg.iterFromRestart = 0
+		}
+	} else {
+		cg.iterFromRestart = 0
```
btracey: And then here.
// It is time to restart. When we restart we take the true gradient as the direction.
(at least, that's what it looks like to me)

btracey:
There need to be tests along with these additions.

cg.go
kortschak: Generally people frown on non-ASCII named vars. I don't think we should use them.

vladimir-ch: I admit that that β variable was just my fancy. With a slight disappointment I will change it to plain beta.

cg.go
vladimir-ch: CGVariant computes the scaling parameter in the direction update formula, which is what differentiates various CG methods one from another. Therefore I like CGVariant, alternatively it could be called CGFormula. CGPreconditioner seems out of place here.

cg.go
vladimir-ch: I basically rewrote NextDirection and added implementation comments explaining the function in detail.

cg.go
```diff
@@ -0,0 +1,293 @@
...
+
+	"github.com/gonum/floats"
+)
+
+const (
+	iterationRestartFactor = 6
+	angleRestartThreshold  = 0.9
+)
+
+// CGVariant represents the formula for computing the scaling parameter (in
```
btracey: "CGVariant calculates the scaling parameter beta" (With greek symbol)

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	Init(loc Location)
+	// Beta returns the value of the scaling parameter that is computed
+	// according to the particular variant of the CG method.
+	Beta(grad, gradPrev, dirPrev []float64) float64
+}
+
+// CG implements the nonlinear conjugate gradient method for solving nonlinear
+// unconstrained optimization problems [1]. It is a line search method that
+// generates the search directions d_k according to the formula
+// d_{k+1} = -∇f_{k+1} + β_k*d_k,   d_0 = -∇f_0.
```
btracey: Space needed for godoc

cg.go
```diff
@@ -0,0 +1,293 @@
...
+// Different variants of the conjugate gradient method differ in the choice
```
btracey: "Variants of ... differ", no need to repeat

cg.go
```diff
@@ -0,0 +1,293 @@
...
+// of the parameter β_k. Conjugate gradient method requires no matrix storage
+// and is faster than the gradient descent method, but L-BFGS is usually more
```
btracey: "Faster" and "more efficient" are misleading, because update flops is one metric,  and function evaluations are a totally different one.  Instead,  just say that CG usually requires fewer function evaluations,  and leave off words that sound like speed. I'm not sure what you mean by lbfgs being more robust, I've heard CG is more robust when the gradients are noisy

cg.go
```diff
@@ -0,0 +1,293 @@
...
+// efficient and robust.
+//
+// CG implements a restart strategy that takes the steepest descent direction
+// (i.e., d_{k+1} = -∇f_{k+1}) whenever any of the following conditions holds.
+// * A certain number of iterations has elapsed without a restart. This number
```
btracey: Again,  spaces are needed for godoc. Make sure you verify that asterisks are okay for bullets (don't have access atm)

jonlawlor: They come through as asterisks just fine.  Also, the ArmijoConditionMet function uses a different format for bullet points ("-" instead of "*") and also has incorrect spacing in its fixed width section.  Several of the comments have different fixed width formatting sections.

You should be able to see it at http://godoc.org/github.com/jonlawlor/optimize#CG

btracey: @jonlawlor  is right, we should keep with the "-". Why is the Armijio spacing wrong? I observe visually that it is, but the actual comment looks basically identical to that of StrongWolfeConditionsMet which shows up in godoc just fine.

cg.go
```diff
@@ -0,0 +1,293 @@
...
+//   depends on the dimension of the problem and is controllable via
```
btracey: Say "This number is controllable via IterationRestartFactor and if 0 is set to a reasonable default based on the problem dimension"

cg.go
```diff
@@ -0,0 +1,293 @@
...
+//   IterationRestartFactor.
+// * The angle between the gradients at two consecutive iterations ∇f_k and
+//   ∇f_{k+1} does not satisfy a certain criterion.
+// * β_k given by Variant.Beta() is equal to zero.
+//
+// The line search for CG must yield step sizes that satisfy the strong Wolfe
+// conditions at every iteration, otherwise the generated search direction
+// might fail to be a descent direction. The line search should be more
+// accurate compared to the case of Newton-like methods, which can be achieved
```
btracey: s/accurate/stringent
compared with those for Newton-like methods

cg.go
```diff
@@ -0,0 +1,293 @@
...
+// by setting the gradient constant in the strong Wolfe conditions to a small
+// value.
+//
+// [1] William Hager, Hongchao Zhang, A survey of nonlinear conjugate gradient
+//     methods. Pacific Journal of Optimization, 2 (2006), pp. 35-58.
+type CG struct {
+	// LinesearchMethod must satisfy the strong Wolfe conditions at every
+	// iteration. If LinesearchMethod == nil, an appropriate default is chosen
```
btracey: Just say ", an appropriate default is chosen".

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	// InitialStep estimates the initial line search step size, because the CG
+	// method does not generate well-scaled search directions.
+	// If InitialStep is nil, an appropriate default is chosen.
+	InitialStep StepSizer
+
+	// The iteration is restarted by taking the steepest descent direction
+	// whenever ceil(IterationRestartFactor*(dimension of the problem))
+	// iterations have elapsed without a restart.
+	// For medium and large-scale problems IterationRestartFactor should be set
+	// to 1 and greater values should be used for low-dimensional problems.
```
btracey: to 1, for low-dimensional problems a larger value should be chosen.

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	// Care must be taken because if the resulting number is 1, the iteration
```
btracey: Note that if IterationRestartFactor == 1, CG becomes identical to gradient descent.

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	// will be restarted at every iteration and the CG method will reduce to
+	// the steepest descent method.
+	// If IterationRestartFactor is 0, it will be set to 6.
+	IterationRestartFactor float64
+	// The method is restarted whenever the gradients at two consecutive
```
btracey: AngleRestartThreshold sets the threshold angle for restart. If the angle between two consecutive gradients is greater than AngleRestartThreshold
  formula
the decent direction is the same as the gradient. A value of AngleRestartThreshold closer to 1 will reduce the number of restarts.

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	restartAfter    int
+	iterFromRestart int
+
+	dirPrev      []float64
+	gradPrev     []float64
+	gradPrevNorm float64
+}
+
+func (cg *CG) Init(loc Location, funcInfo *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
+	if cg.LinesearchMethod == nil {
```
btracey: Should there be panics for Angle / IterationRestart being less than zero?

cg.go
```diff
@@ -0,0 +1,293 @@
...
+
+func (cg *CG) NextDirection(loc Location, direction []float64) (stepSize float64) {
+	copy(direction, loc.Gradient)
+	floats.Scale(-1, direction)
+
+	cg.iterFromRestart++
+	var restart bool
+	// Check if the method should be restarted because too many iterations have
+	// been taken without a restart.
+	if cg.iterFromRestart%cg.restartAfter == 0 {
```
btracey: Can we just make it a comparison rather than a mod?

cg.go
btracey: We also need to note that AngleRestartThreshold is the cosine of the angle, not the angle iteslf

cg.go
```diff
@@ -0,0 +1,293 @@
...
+		restart = true
+	}
+	// Check if the method should be restarted because the last two gradients
+	// point in almost opposite directions.
```
btracey: almost orthogonal directions. (based on my understanding of the function)

cg.go
jonlawlor: I think https://github.com/gonum/optimize/blob/master/linesearch.go#L118 has an extra tab.

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	gNorm := floats.Norm(loc.Gradient, 2)
+	gDotGPrev := floats.Dot(loc.Gradient, cg.gradPrev)
```
btracey: can we just call this gDot?

cg.go
btracey: Weird ... why don't I see that locally?

````
brendan:~/Documents/mygo/src/github.com/gonum/optimize$ git branch
  bfgsinit
  decreasetol
* master
brendan:~/Documents/mygo/src/github.com/gonum/optimize$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

nothing to commit, working directory clean
brendan:~/Documents/mygo/src/github.com/gonum/optimize$ git pull
Already up-to-date.
````

File: cg.go [hunk 22]
```diff
@@ -0,0 +1,293 @@
...
+	gNorm := floats.Norm(grad, 2)
+	beta = gNorm * gNorm / fr.gPrevNorm / fr.gPrevNorm
+
+	fr.gPrevNorm = gNorm
+	return beta
+}
+
+// PolakRibierePolyak implements the Polak-Ribiere-Polyak variant of the CG
+// method that computes the scaling parameter β_k according to the formula
+// β_k = max(0, ∇f_{k+1}·y_k / |∇f_k|^2),
```
btracey: Formula should be indented with a space so it gets its formatting box. This should happen with all of the variants for both of the formulas.

cg.go
jonlawlor: Maybe your text editor is hiding it?  Try
```bash
cat linesearch.go | grep -m 1 initGrad
// 	- initGrad < 0
```

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	gPrevNorm float64
+}
+
+func (fr *FletcherReeves) Init(loc Location) {
+	fr.gPrevNorm = floats.Norm(loc.Gradient, 2)
+}
+
+func (fr *FletcherReeves) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	gNorm := floats.Norm(grad, 2)
+	beta = gNorm * gNorm / fr.gPrevNorm / fr.gPrevNorm
```
btracey: Please make it
````
beta = (gNorm / fr.gPrevNorm) * (gNorm / gPrevNorm)
````
Should be better behaved in edge cases.

File: cg.go [hunk 24]
```diff
@@ -0,0 +1,293 @@
...
+	copy(cg.dirPrev, direction)
+	cg.gradPrevNorm = gNorm
+	return stepSize
+}
+
+// FletcherReeves implements the Fletcher-Reeves variant of the CG method that
+// computes the scaling parameter β_k according to the formula
+// β_k = |∇f_{k+1}|^2 / |∇f_k|^2.
+type FletcherReeves struct {
+	gPrevNorm float64
```
btracey: Just call it prevNorm. There's no other norm or other variables involved at all

cg.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func (fr *FletcherReeves) Init(loc Location) {
+	fr.gPrevNorm = floats.Norm(loc.Gradient, 2)
+}
+
+func (fr *FletcherReeves) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	gNorm := floats.Norm(grad, 2)
```
btracey: Just call it norm (see above)

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	fr.gPrevNorm = gNorm
+	return beta
+}
+
+// PolakRibierePolyak implements the Polak-Ribiere-Polyak variant of the CG
+// method that computes the scaling parameter β_k according to the formula
+// β_k = max(0, ∇f_{k+1}·y_k / |∇f_k|^2),
+// where y_k = ∇f_{k+1} - ∇f_k.
+type PolakRibierePolyak struct {
+	gPrevNorm float64
```
btracey: Same comments about norm

cg.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func (pr *PolakRibierePolyak) Init(loc Location) {
+	pr.gPrevNorm = floats.Norm(loc.Gradient, 2)
+}
+
+func (pr *PolakRibierePolyak) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	gNorm := floats.Norm(grad, 2)
+	gDotGPrev := floats.Dot(grad, gradPrev)
```
btracey: You can call this "dot"

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	beta = (gNorm*gNorm - gDotGPrev) / pr.gPrevNorm / pr.gPrevNorm
```
btracey: This is probably better separated out too
beta = (norm / pr.prevNorm) * (norm / pr.prevNorm) - dot / (pr.prevNorm * pr.prevNorm)

btracey: Actually, I take it back. probably best as
beta = (norm * norm - dot) / (pr.prevNorm * pr.prevNorm)

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	y []float64
+}
+
+func (hz *HagerZhang) Init(loc Location) {
+	hz.y = resize(hz.y, len(loc.Gradient))
+}
+
+func (hz *HagerZhang) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	floats.SubTo(hz.y, grad, gradPrev)
+	dDotY := floats.Dot(dirPrev, hz.y)
```
btracey: dirDotY

cg.go
```diff
@@ -0,0 +1,293 @@
...
+	gDotY := floats.Dot(grad, hz.y)
+	gDotD := floats.Dot(grad, dirPrev)
```
btracey: gDotDir

cg.go
vladimir-ch: It would be *far from* orthogonal directions but only if the dot product were in an absolute value. As it is now, it indeed restarts as it is written, that is, when the gradients point in almost opposite directions. I chose the non-absolute-value version as it seems to assure convergence in the same number of cases with less restarts.

cg.go
btracey: Err, cos of the angle.

cg.go
btracey: Yea, you're right, I got confused with the negative sign. My fault.

Would it make more sense to have the restart happen if the angle is less than AngleThreshhold and remove the negative sign? In this case, AngleThreshhold would default to -0.9. This way it aligns with the Cos formula, 1 and they're perfectly aligned, 0 they're othogonal, -1 they're exactly opposite.

jonlawlor:
I believe https://github.com/gonum/optimize/blob/master/unconstrained_test.go#L11 needs to be changed.

vladimir-ch:
I think I addressed all your comments, @btracey , could you PTAL? I am not sure if the comments for AngleRestartFactor and IterationsRestartFactor are clear enough. CG panics if AngleRestartFactor is not in [-1,0], because that is the only reasonable interval of values for it, but it too can be reconsidered.

vladimir-ch:
Also, I will add tests later after reviewing and merging change-tests branch.

cg.go
```diff
@@ -0,0 +1,317 @@
...
+// L-BFGS is usually more efficient.
+//
+// CG implements a restart strategy that takes the steepest descent direction
+// (i.e., d_{k+1} = -∇f_{k+1}) whenever any of the following conditions holds:
+//
+//   - A certain number of iterations has elapsed without a restart. This number
+//     is controllable via IterationRestartFactor and if equal to 0, it is set to
+//     a reasonable default based on the problem dimension.
+//   - The angle between the gradients at two consecutive iterations ∇f_k and
+//     ∇f_{k+1} does not satisfy a certain criterion.
```
btracey: instead of "does not satisfy a certain criterion" can we just say "is too large"?

cg.go
```diff
@@ -0,0 +1,317 @@
...
+//   - β_k given by Variant.Beta() is equal to zero.
```
btracey: CGVaraint.Beta()

cg.go
```diff
@@ -0,0 +1,317 @@
...
+	// IterationRestartFactor should be set to 1, low-dimensional problems a
+	// larger value should be chosen. Note that if the ceil function returns 1,
+	// CG will be identical to gradient descent.
+	// If IterationRestartFactor is 0, it will be set to 6.
+	// CG will panic if IterationRestartFactor is negative.
+	IterationRestartFactor float64
+	// AngleRestartThreshold sets the threshold angle for restart. If the
+	// cosine of the angle between two consecutive gradients is smaller than
+	// AngleRestartThreshold, that is, if
+	//
```
btracey: The extra newlines aren't necessary.

cg.go
```diff
@@ -0,0 +1,317 @@
...
+	//   ∇f_k·∇f_{k+1} / (|∇f_k| |∇f_{k+1}|) < AngleRestartThreshold,
+	//
+	// the descent direction is the same as the negative of the gradient.
```
btracey: What does this comment have to do with anything? The descent direction is not mentioned anywhere else.

I think it's fine to just delete this sentence and end the thought at the comment.

cg.go
```diff
@@ -0,0 +1,317 @@
...
+	// A value of AngleRestartThreshold closer to -1 will reduce the number of
```
btracey: "closer to -1 (successive gradients in exact opposite directions) will tend to reduce"

btracey:
I had a couple of minor comments, but nice job on the documentation.

cg.go
```diff
@@ -0,0 +1,317 @@
...
+	// Check if the method should be restarted because too many iterations have
+	// been taken without a restart.
+	if cg.iterFromRestart == cg.restartAfter {
+		restart = true
+	}
+	// Check if the method should be restarted because the angle between
+	// the last two gradients is too large.
+	gDot := floats.Dot(loc.Gradient, cg.gradPrev)
+	gNorm := floats.Norm(loc.Gradient, 2)
+	cos := math.Max(-1, gDot/(gNorm*cg.gradPrevNorm))
```
btracey: What does the Max do? It seems to me like the only effect is has is to allow AngleRestartThreshold == -1 to act as "never restart because of the angle" even when the gradients are really poorly scaled. If the line below this was "<=" instead of less than, the max would have no effect, because cg.AngleRestartThreshold >= -1.

cg.go
```diff
@@ -0,0 +1,317 @@
...
+	if cos < cg.AngleRestartThreshold {
+		restart = true
+	}
+
+	// Compute the scaling factor β_k according to the given CG variant. Do
+	// this always, because cg.Variant may be keeping its inner state that
```
btracey: s/alwals/even when restarting

cg.Variant may be keeping _an_ inner state  (underscores mine to highlight the change, don't actually include them)

cg.go
```diff
@@ -0,0 +1,317 @@
...
+
+	copy(cg.gradPrev, loc.Gradient)
+	copy(cg.dirPrev, direction)
+	cg.gradPrevNorm = gNorm
+	return stepSize
+}
+
+// FletcherReeves implements the Fletcher-Reeves variant of the CG method that
+// computes the scaling parameter β_k according to the formula
+//
```
btracey: extra line not needed. This is true for all below.

cg.go
```diff
@@ -0,0 +1,317 @@
...
+//   β_k = |∇f_{k+1}|^2 / |∇f_k|^2.
```
btracey: You only need one extra space. (below as well)

cg.go
```diff
@@ -0,0 +1,317 @@
...
+
+func (pr *PolakRibierePolyak) Init(loc Location) {
+	pr.prevNorm = floats.Norm(loc.Gradient, 2)
+}
+
+func (pr *PolakRibierePolyak) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	norm := floats.Norm(grad, 2)
+	dot := floats.Dot(grad, gradPrev)
+	beta = (norm*norm - dot) / (pr.prevNorm * pr.prevNorm)
+
```
btracey: remove newline.

cg.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func (fr *FletcherReeves) Init(loc Location) {
+	fr.prevNorm = floats.Norm(loc.Gradient, 2)
+}
+
+func (fr *FletcherReeves) Beta(grad, gradPrev, dirPrev []float64) (beta float64) {
+	norm := floats.Norm(grad, 2)
+	beta = (norm / fr.prevNorm) * (norm / fr.prevNorm)
+
```
btracey: remove newline

btracey:
LGTM when the comments are addressed. The only comment that addresses code function is the Max question. The rest are all documentation / minor style.

Thanks for working with me. This all looks really nice now.

cg.go
vladimir-ch: Yes, allowing AngleRestartThreshold == -1 to skip the angle test was the intention. Given that CG easily fails without this condition, it make no sense to try to reserve -1 for skipping it. I modified the condition (and the docs) back to <= and removed the Max.

vladimir-ch:
Thanks for reviewing the code, @btracey . Writing a compact, clear Go style documentation is hard, your comments have really helped me to improve it (not to mention the code itself).

PR #23, opened 2014-12-17
Add Shanno-Phua initial step size

vladimir-ch:
* Implemented the other initial step size strategy described in Nocedal&Wright
* In some CG implementations the formula is attributed to Shanno&Phua, but for consistence with QuadraticStepSize the type's name FirstOrderStepSize rather refers to how the step size is estimated. ShannoPhuaStepSize is an alternative name anyway.
* This strategy seems to generate step size that work better with CG than those generated by QuadraticStepSize
* #16 should be reviewed and merged first

### DIFF
--- a/stepsizers.go
+++ b/stepsizers.go
@@ -41,19 +41,20 @@ func (c ConstantStepSize) StepSize(l Location, dir []float64) float64 {
 // descent directions, such as gradient descent or conjugate gradient methods.
 // The step size is bounded away from zero.
 type QuadraticStepSize struct {
-	// If the relative change in the objective function is larger than
-	// Threshold, the step size is estimated by quadratic interpolation,
-	// otherwise it is set to 2*previous step size.
-	// The default value is 1e-12.
+	// Threshold determines that the initial step size should be estimated by
+	// quadratic interpolation when the relative change in the objective
+	// function is larger than Threshold.  Otherwise the initial step size is
+	// set to 2*previous step size.
+	// If Threshold is zero, it will be set to 1e-12.
 	Threshold float64
-	// The step size at the first iteration is estimated as InitialStepFactor / |g|_∞.
+	// InitialStepFactor sets the step size for the first iteration to be InitialStepFactor / |g|_∞.
 	// If InitialStepFactor is zero, it will be set to one.
 	InitialStepFactor float64
-	// The estimated step size is always greater than or equal to MinStepSize.
+	// MinStepSize is the lower bound on the estimated step size.
 	// MinStepSize times GradientAbsTol should always be greater than machine epsilon.
 	// If MinStepSize is zero, it will be set to 1e-3.
 	MinStepSize float64
-	// The estimated step size is always lower than or equal to MaxStepSize.
+	// MaxStepSize is the upper bound on the estimated step size.
 	// If MaxStepSize is zero, it will be set to 1.01.
 	MaxStepSize float64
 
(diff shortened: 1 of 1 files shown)


stepsizers.go
```diff
@@ -27,63 +36,136 @@ func (c ConstantStepSize) StepSize(l Location, dir []float64) float64 {
...
+	stepSize = math.Max(q.MinStepSize, math.Min(stepSize, q.MaxStepSize))
 
 	q.fPrev = l.F
-	return stepsize
+	q.dirPrevNorm = floats.Norm(dir, 2)
+	q.projGradPrev = projGrad
+	copy(q.xPrev, l.X)
+	return stepSize
+}
+
```
btracey: This needs a high-level comment about what it does.

vladimir-ch:
PTAL @btracey

stepsizers.go
```diff
@@ -115,3 +118,61 @@ func (q *QuadraticStepSize) StepSize(l Location, dir []float64) (stepSize float6
...
+// FirstOrderStepSize estimates the initial line search step size based on the
+// assumption that the first-order change in the function will be the same as
+// that obtained at the previous iteration. That is, the initial step size s^0_k
+// is chosen so that
+//   s^0_k ∇f_k⋅p_k = s_{k-1} ∇f_{k-1}⋅p_{k-1}
+// This is useful for line search methods that do not produce well-scaled
+// descent directions, such as gradient descent or conjugate gradient methods.
+type FirstOrderStepSize struct {
+	// The step size at the first iteration is estimated as InitialStepFactor / |g|_∞.
```
btracey: This is backward. 
... estimated at |g|_inf / InitialStepFactor

stepsizers.go
```diff
@@ -115,3 +118,61 @@ func (q *QuadraticStepSize) StepSize(l Location, dir []float64) (stepSize float6
...
 }
+
+// FirstOrderStepSize estimates the initial line search step size based on the
+// assumption that the first-order change in the function will be the same as
+// that obtained at the previous iteration. That is, the initial step size s^0_k
+// is chosen so that
+//   s^0_k ∇f_k⋅p_k = s_{k-1} ∇f_{k-1}⋅p_{k-1}
+// This is useful for line search methods that do not produce well-scaled
+// descent directions, such as gradient descent or conjugate gradient methods.
+type FirstOrderStepSize struct {
```
btracey: Reading these, I think it would be nicer if the struct comments follow the Go style. "Initial step factor sets the step size for the first iteration to be |g|_inf / InitialStepFactor"

I know this isn't consistent at the moment, but reading these for something I don't know makes me realize I've gotten used to reading godoc style comments.

stepsizers.go
```diff
@@ -115,3 +118,61 @@ func (q *QuadraticStepSize) StepSize(l Location, dir []float64) (stepSize float6
...
+	fo.xPrev = resize(fo.xPrev, len(l.X))
+	if fo.InitialStepFactor == 0 {
+		fo.InitialStepFactor = initialStepFactor
+	}
+	if fo.MinStepSize == 0 {
+		fo.MinStepSize = firstOrderMinimumStepSize
+	}
+	if fo.MaxStepSize == 0 {
+		fo.MaxStepSize = firstOrderMaximumStepSize
+	}
```
btracey: We should probably have a check / panic if MaxStep < MinStep

btracey:
LGTM with the comment changes.

stepsizers.go
vladimir-ch: This is correct, we want to take a small step when the gradient is large (when it is small, the step size is bound by MaxStepSize).

stepsizers.go
vladimir-ch: I try to follow godoc style everywhere, here it somehow escaped my attention. Fixed.

stepsizers.go
vladimir-ch: Added.

PR #24, opened 2015-01-09
Fix some golint warnings

vladimir-ch:
* Mistyped type names in comments
* Dots at the end of error strings
* Replace -= 1 with --


kortschak:
LGTM

PR #28, opened 2015-01-15
Reorganize and add unconstrained tests

vladimir-ch:


types.go
```diff
@@ -9,8 +9,10 @@ import (
 	"time"
 )
 
-// EvaluationType is used by the optimizer to specify information needed from
-// the objective function.
+const defaultGradientAbsTol = 1e-6
+
+// EvalutationType is used by the optimizer to specify information needed
```
kortschak: s/EvalutationType/EvaluationType/

unconstrained_test.go
```diff
@@ -884,16 +915,286 @@ func (Linear) Df(x, grad []float64) {
 	}
 }
 
+type UnconstrainedTest struct {
+	// f is the function that is being minimized
```
kortschak: Full stops on comments.

unconstrained_test.go
```diff
@@ -901,130 +1202,115 @@ func TestGradientDescentBacktracking(t *testing.T) {
...
-				},
-			},
-		*/
-	} {
-		test.Settings.Recorder = nil
-		result, err := Local(test.F, test.X, test.Settings, method)
+		result, err := Local(test.f, test.x, settings, method)
 		if err != nil {
-			t.Errorf("error finding minimum: %v", err.Error())
+			t.Errorf("error finding minimum: %v\n%v", err.Error(), test)
```
kortschak: Drop .Error() method call when using %v with errors.

unconstrained_test.go
```diff
@@ -901,130 +1202,115 @@ func TestGradientDescentBacktracking(t *testing.T) {
...
+			settings.InitialFunctionValue = funcs.gradFunc.FDf(test.x, settings.InitialGradient)
+			settings.UseInitialData = true
+		}
+
+		// Rerun the test again to make sure that it gets the same answer with
+		// the same starting condition.
+		result2, err2 := Local(test.f, test.x, settings, method)
 		if err2 != nil {
-			t.Errorf("error finding minimum second time: %v", err2.Error())
+			t.Errorf("error finding minimum second time %v for:\n%v", err2.Error(), test)
```
kortschak: As above.

unconstrained_test.go
```diff
@@ -884,16 +915,286 @@ func (Linear) Df(x, grad []float64) {
...
+	gradTol float64
+	// tol is the tolerance for checking the accuracy of result.F. If tol == 0,
+	// the default tolerance will be used.
+	tol float64
+}
+
+func (t UnconstrainedTest) String() string {
+	dim := len(t.x)
+	if dim <= 10 {
+		// Print the initial and optimal X only for small-dimensional problems
```
kortschak: Full stop.

unconstrained_test.go
```diff
@@ -901,130 +1202,115 @@ func TestGradientDescentBacktracking(t *testing.T) {
...
 			}
-		*/
+		} else if funcInfo.IsFunctionGradient {
+			if result.FunctionGradientEvals != result2.FunctionGradientEvals+1 {
+				t.Errorf("Providing initial data does not reduce the number of function/gradient calls for:\n%v", test)
+				continue
+			}
+		}
+
+		// if test.optLoc != nil {
```
kortschak: Why is this commented out?

vladimir-ch: Optimizing from a (nearly) optimal location should successfully return either immediately or after a few iterations. However, due to the floating point effects the current implementation cannot handle many of the tests, either an error is returned or the call to Local does not even return. This is serious and must be addressed somehow, so this is commented out to serve as a kind of TODO.

kortschak: Add an explicit TODO?

unconstrained_test.go
```diff
@@ -884,16 +915,286 @@ func (Linear) Df(x, grad []float64) {
 	}
 }
 
+type UnconstrainedTest struct {
+	// f is the function that is being minimized.
+	f Function
+	// x is the initial approximation.
```
btracey: initial guess

unconstrained_test.go
```diff
@@ -884,16 +915,286 @@ func (Linear) Df(x, grad []float64) {
...
+	},
+	{
+		f:      Wood{},
+		x:      []float64{-3, -1, -3, -1},
+		optVal: 0,
+		optLoc: []float64{1, 1, 1, 1},
+	},
+}
+
+var bfgsTests = []UnconstrainedTest{
```
btracey: Could you add some comment to these as to what the intended purpose is? I can see that gradient tests are for all gradient-based optimizers, newton tests are for newtons, but it's not clear to me why there needs to be the third level of bfgs and lbfgs tests.

unconstrained_test.go
```diff
@@ -901,130 +1202,117 @@ func TestGradientDescentBacktracking(t *testing.T) {
...
 		}
-		if result == nil {
-			t.Errorf("Case %v: nil result without error", i)
+
+		// Check that the optimum function value is as expected.
+		if math.Abs(result.F-test.optVal) > test.tol {
+			t.Errorf("Minimum not found, exited with status: %v. Want: %v, Got: %v for:\n%v",
+				result.Status, test.optVal, result.F, test)
 			continue
 		}
```
btracey: Could you please add a test here which evaluates the function (test.f) at the found optimum location and ensures that the gradient is lower than the tolerance?

unconstrained_test.go
```diff
@@ -901,130 +1202,117 @@ func TestGradientDescentBacktracking(t *testing.T) {
...
 
-		// rerun it again to ensure it gets the same answer with the same starting
-		// condition
-		result2, err2 := Local(test.F, test.X, test.Settings, method)
+		funcs, funcInfo := getFunctionInfo(test.f)
```
btracey: Could you add a comment that we are going to restart the solution using a fixed starting gradient and value.

unconstrained_test.go
```diff
@@ -901,130 +1202,117 @@ func TestGradientDescentBacktracking(t *testing.T) {
...
-		*/
+			if funcInfo.IsGradient {
+				if result.GradientEvals != result2.GradientEvals+1 {
+					t.Errorf("Providing initial data does not reduce the number of gradient calls for:\n%v", test)
+					continue
+				}
+			}
+		}
+
+		// if test.optLoc != nil {
```
btracey: In addition to the TODO mentioned by @kortschak, please add a comment saying "Try starting the optimizer from a (nearly) optimum condition. It's reasonably clear from the test error message, but it's easy to miss the "test.optLoc" in the Local call.

unconstrained_test.go
vladimir-ch: There are problems that BFGS and LBFGS handle equally well, but then there are problems that one method handles better than the other (e.g., can achieve lower gradient norm, converges from a point much further from the minimum, ...). The idea was to put the former problems into newtonTests and the former into bfgsTests or lbfgsTests - if a method can solve a more difficult problem, then why not include it in the tests. I was not sure about it either and I hoped that it would come up during the review.

This touches a little on the issue of the purpose of these tests - are we testing for correctness of the implementation, for the ability to solve as hard problems as possible (badly scaled, stringent tolerances, ...) or some other criteria not yet included?

vladimir-ch: errr, the former ... and the latter ...

unconstrained_test.go
vladimir-ch: Sure, good idea.

unconstrained_test.go
vladimir-ch: I can make the tests in (l)bfgsTests easier and move them to newtonTests and remove (l)bfgsTests. When CG is added, there would be another group of tests (many of newtonTests could be moved there). GradientDescent would be tested with gradientDescentTests, CG with gradientDescentTests and cgTests and (L)BFGS moreover with newtonTests. The purpose is to avoid defining identical/very similar tests for each method.

unconstrained_test.go
vladimir-ch: Hm, interesting. I added the test, but surprisingly it fails. Either the test is wrong and I don't see it or there is indeed something wrong with the implementation (and I don't see it).

result.GradientNorm _is_ smaller than the tolerance, but the norm that I evaluate at result.X is not equal to result.GradientNorm and is greater than the tolerance. Could you PTAL, @btracey ?

vladimir-ch: Except for this, I think I addressed all the comments.

vladimir-ch: It seems that there is some kind of floating point effect taking place. The test was indeed failing, but only for some test problems, the most difficult ones (badly scaled, minimum at the bottom of an almost flat plateau). Loosening the GradientAbsTol passes the test. However, in one test problem I had to increase the tolerance up to 1e-4. Also, I still don't understand where the result.Gradient and result.GradientNorm come from in case the test fails.

Moreover, `go test` passes on my computer, but apparently fails on Travis. I can increase the tolerance further, but how to approach this in a more reliable way? Any suggestions, @btracey @kortschak ?

btracey: When you say "the test" you're referring that the gradient matches? Could you post the code segment you inserted? You're using the infinity norm and not the 2-norm?

btracey: Sorry, I see it now. With the code as is, go test fails on my computer. Looking into it.

vladimir-ch: The easiest solution would be to give up stringent tolerances in the tests and use the default tolerance (1e-6) everywhere except those few problems that need 1e-5 or 1e-4. This comes back to the question of what we are testing: correctness of the code, ability to optimize with the default tolerance or ability to optimize with tolerances pushed to the limits?

vladimir-ch: The three problematic tests between lines 1292-1309.

Another possibility is that there could be a mistake in some of the test functions, although I tried to make sure that everything is ok. I will check them.

btracey: You should verify. I notice that none of the FunctionGradient tests fail. Not sure if that's a coincidence, but maybe we have an update wrong somewhere.

btracey: I think we are testing correctness of the code, which means limits higher than the default. I would love to get machine-precision, but because of floating-point issues that's really hard (though it's why there is the floating point check in bisection). In all cases, getting to within 1e-10 should be doable, and 1e-12 would be better. If there's a good explanation why we can't get tighter then it's one thing, but while we're still trying to confirm correctness, I'm nervous of lifting it too much. Floating point error is significant, but not so enormous we can't find good optima (and the optimizer should definitely agree on the function value and gradient at optimum)

btracey: Here's a possible problem, though it doesn't seem to fix the tests. Right now in update, we have
````
if location.F < optLoc.F {
    copyLocation(optLoc, location)
}
````

Update is called on every function evaluation, but the evaluations are sometimes of the function, sometimes for the gradient. For the cases where it's two-stage, I think the location would be copied the first time through. It should probably be a "<=". 

We may need to do the check here we do in bisection. I can try and look into it.

Comment (vladimir-ch) [inline unconstrained_test.go]: > I notice that none of the FunctionGradient tests fail. Not sure if that's a coincidence, but maybe we have an update wrong somewhere.

I implemented FDf() only for the Rosenbrock function, that's a possible reason.

I was just about to post the same possible problem. I see also another possible one. A few lines below we do:
```
if location.Gradient != nil {
	stats.GradientNorm = floats.Norm(location.Gradient, math.Inf(1))
}
```
This could make stats.GradientNorm out of sync with the gradient in optLoc.

Comment (vladimir-ch) [inline unconstrained_test.go]: If I modify the code so that it is always in sync, I no longer get the failure because result.GradientNorm does not matches the one evaluated at result.X, instead I get a normal failure from the line search.

Comment (vladimir-ch) [inline unconstrained_test.go]: What about moving GradientNorm from stats to Location? It would be updated in evaluate() and thus always in sync.

Comment (btracey) [inline unconstrained_test.go]: > On Jan 15, 2015, at 10:06 PM, Vladimír Chalupecký <notifications@github.com> wrote:
> 
> In unconstrained_test.go <https://github.com/gonum/optimize/pull/28#discussion-diff-23064677>:
> 
> >  			continue
> >  		}
> > -		if result == nil {
> > -			t.Errorf("Case %v: nil result without error", i)
> > +
> > +		// Check that the optimum function value is as expected.
> > +		if math.Abs(result.F-test.optVal) > test.tol {
> > +			t.Errorf("Minimum not found, exited with status: %v. Want: %v, Got: %v for:\n%v",
> > +				result.Status, test.optVal, result.F, test)
> >  			continue
> >  		}
> If I modify the code so that it is always in sync, I no longer get the failure because result.GradientNorm does not matches the one evaluated at result.X, instead I get a normal failure from the line search.
> 
Great! Could you push that change so I can help debug the line search?

Comment (vladimir-ch) [inline unconstrained_test.go]: > Great! Could you push that change so I can help debug the line search?

The modification was just a quick hack. Do you agree with moving GradientNorm to Location or do you prefer another solution?

Comment (btracey) [inline unconstrained_test.go]: There may be some floating point oddities that caused that out-of-sync. Places where floating point noise means the function value is slightly higher even though the gradient norm has actually decreased? Is it something simpler than that that I'm missing?

Comment (btracey) [inline unconstrained_test.go]: All things being equal, I'd prefer it stay out, but if it's causing a bunch of trouble I could bend. There are two issues with including it:
1) Location is used for the starting location, and we don't want the user (having to) fill in the information about the gradient. It would be nicer if it was invisible completely, which means making it private
2) If it's private, we need to make sure our optimizers don't actually use it. I like that our optimizers are not privileged -- anyone could re-implement externally the optimizers we have and plug them into Local.

It's not clear to me what the problem is yet and how that's solved by moving the gradient storage location.

Comment (vladimir-ch) [inline unconstrained_test.go]: I think you are right, it will be something like that (floating point oddities).

Also, if there were a FunctionRelTol, it might signal convergence before these floating point oddities arise. But that's for another PR.

Having it in Location would mean that if loc.Gradient != nil, loc.GradientNorm always corresponds to loc.Gradient because evaluate() would . Currently, the problem is that

1. stats.GradientNorm corresponds to loc.Gradient, not to optLoc which is the one returned from Local()
2. convergence is checked against stats.GradientNorm (i.e., |loc.Gradient| )

Perhaps a final copyLocation() before returning from Local() would help.

Comment (vladimir-ch) [inline unconstrained_test.go]: I take the last sentence back ... I don't know how to phrase it precisely, we should return the location that triggers the convergence, not optLoc because that is behind the location.F <= optLoc.F filter.

Comment (vladimir-ch) [inline unconstrained_test.go]: Also, Location _is_ used for the starting location, but it is not filled in by the user. The user sees Location only in Result.

Furthermore, there is no point in hiding GradientNorm from the optimizers, they already have access to Location.Gradient. I don't see any harm in having GradientNorm in Location, quite the opposite.

Comment (btracey) [inline unconstrained_test.go]: Yep, you're right. I had thought we had Location in settings, but we don't. Find to add, and the change to a pointer is a good one.

Comment (btracey) [inline unconstrained_test.go]: OptLoc should be the one returned and the one tested for convergence. We can't just test the most recent location, because individual function evaluations may be worse, and we treat the Function tolerance as a strict cap. I think we can get away with a more nuanced test than "<=", for example what we do in Bisection. as long as we're testing the right thing for convergence, we should be consistent.

**Comment by vladimir-ch** [comment]
OK, coming back to this PR which had revealed some issues with the then implementation. It triggered a series of commits (e7e281fa685ef5026458da5e0b4405364b1f8fe8..1ede5c9909eb173549ae696475fe6472f0accb2a)  that fixed them (AFAICT). Except for a few cases, most tests pass with GradientAbsTol = 1e-12 (including Travis). There is still the outstanding issue of optimizing from (nearly) optimal location, but let's leave that for another PR.

Could you PTAL, @btracey ?

**Comment by btracey** [comment]
Before we do this, can we fix #53 ? It may fix the nearly optimal location, and it'll let us know how far we can optimize. I would guess it's still about 1e-12, but it would be good to check. This way these tests can go in cleanly. I can put that together if you'd like, otherwise I don't want to stop your warpath if you'd like to fix it.

**Comment by vladimir-ch** [comment]
Sure, go ahead, I didn't intend to modify your Bisection.

A warpath? :-) I just wanted to fix a problem before I forgot what it was.

**Comment by btracey** [comment]
I use that term positively. It's how I would have described my work on BLAS around the new year.

**Comment by vladimir-ch** [comment]
> I use that term positively.

I know. :-)

**Comment by btracey** [comment]
Do things still look good with that fix?

File: unconstrained_test.go [hunk 12]
```diff
@@ -884,16 +915,302 @@ func (Linear) Df(x, grad []float64) {
...
+	}
+}
+
 func TestMinimize(t *testing.T) {
-	testMinimize(t, nil)
+	// TODO: When method is nil, Local chooses the method automatically. At
+	// present, it always chooses BFGS (or panics if the function does not
+	// implement Df() or FDf()). For now, run this test with the simplest set
+	// of problems and revisit this later when more methods are added.
+	// Alternatively, we could remove (or better, change) this test because
```
btracey: No, this test is here to make sure the full integrated test with Local works when the optimizer is unspecified.  We can add TestGetDefaultMethod if we want to check that does the right thing. This test here is a full-integration test, and given how common the call will be, it's worth having even if it is redundant with other tests.

vladimir-ch: OK, I will remove the second half of the comment.

vladimir-ch:
I had to loosen the gradient tolerance on some notoriously problematic functions, but other than that it still looks good. As for the optimization from nearly optimum location, there is no change, the test does not finish.

btracey:
It fixes the issue of the gradient being larger than the specified tolerance though, right?

btracey:
LGTM

vladimir-ch:
Yes, that issue is fixed.

PR #34, opened 2015-01-17
Remove optloc

btracey:


local.go
```diff
@@ -196,46 +197,40 @@ func getDefaultMethod(funcInfo *FunctionInfo) Method {
 // Combine location and stats because maybe in the future we'll add evaluation times
 // to functionStats?
 func getStartingLocation(f Function, funcs functions, funcInfo *FunctionInfo, initX []float64, stats *Stats, settings *Settings) (Location, error) {
```
btracey: This should just return a *Location

local.go
```diff
@@ -83,40 +82,42 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
 	// cleanup at exit
 	if settings.Recorder != nil && err == nil {
-		err = settings.Recorder.Record(*optLoc, NoEvaluation, PostIteration, stats)
+		err = settings.Recorder.Record(optLoc, NoEvaluation, PostIteration, stats)
 	}
 	stats.Runtime = time.Since(startTime)
 	return &Result{
 		Stats:    *stats,
-		Location: *optLoc,
+		Location: optLoc,
```
btracey: This should probably promoted to a pointer too. We haven't quite answered how we're going to deal with constraints, but if they go in Location the struct will become too big to just copy everywhere. Making it a pointer seems for the best.

btracey: (Which includes changing the interface types I think)

local.go
```diff
@@ -328,32 +334,9 @@ func evaluate(funcs functions, funcInfo *FunctionInfo, evalType EvaluationType,
 }
 
 // update updates the stats given the new evaluation
```
btracey: It seems like at this point, update can just go away. I had originally wanted to separate with a future Global, but it'll probably need it's own function anyway.

local.go
```diff
@@ -196,46 +197,40 @@ func getDefaultMethod(funcInfo *FunctionInfo) Method {
...
 // to functionStats?
 func getStartingLocation(f Function, funcs functions, funcInfo *FunctionInfo, initX []float64, stats *Stats, settings *Settings) (Location, error) {
-	var l Location
-
-	l.X = make([]float64, len(initX))
-	copy(l.X, initX)
-	if funcInfo.IsGradient || funcInfo.IsFunctionGradient {
-		l.Gradient = make([]float64, len(l.X))
+	dim := len(initX)
+	loc := Location{
```
btracey: Would you mind changing the various "loc" back to their original name? I'm okay with fixing them, but it obscures this diff quite a bit. We can fix it in a future PR.

btracey:
I think we need to keep optLoc. It prevents two cases. 
1) The optimizer accidentally jumps to a relative maximum. Function value increases, but the gradient norm is small.
2) The drastically more common case where a function evaluation limit is reached in the middle of a linesearch. We should return the lowest value found, not the most recent function execution. 

The trick is keeping OptLoc in sync with what the optimizer thinks is the best location. This is easy except for floating point noise that makes a direct function value comparison throw off the optimizer really close to the optimum. I think there are two solutions
1) Take the comparison style that's in Bisection,  turn it into a real function, and call it both in evaluate/update and in Bisection. It should probably be exported so that outside optimizers can use it. (See addendum below)
2) Use the explicit function comparison, and sacrifice some ability  to converge the gradient really low. This is easier to understand, simpler to code, and less likely to have unintended consequences. Additionally, there are free parameters in the function comparison, which are a tradeoffs between floating point noise and actual function oscillation. As functions get larger and more complicated, the floating point noise increases, so it's not clear that one can even construct reasonable defaults for the parameters. I've run into this issue in my own functions, especially when concurrency is involved. I think users could run an additional minimization on the norm of the gradient for those who want increased convergence (by changing the objective function and starting from the previously found minimum). We could plausibly provide support for this with a provided objective function.

In either case, this merits some commentary in the documentation.

local.go
vladimir-ch: I agree with changing all Location to *Location. It is a big change, though, so I propose to split this PR into several smaller ones.

local.go
vladimir-ch: That's exactly what I thought, but let's first make the other changes.

vladimir-ch:
Yes, I agree, removing optLoc was a too drastic idea, let's keep it. But:
1) For line search methods, the function value should never increase thanks to Armijo (talking about Armijo, why is FunConst in Bisection equal to 0?). Trust region also guarantees monotonous decrease in function values. I am not so sure about Nelder-Mead (or non-deterministic methods). The question is, who should take care of keeping track of the optimum location? minimize() or Method? I am slightly leaning towards Method. In Linesearch we require NextDirectioner to return a descent direction, otherwise we return an error. Requiring LinesearchMethod to return steps that decrease function value between two MajorIteration is the same class of requirement. 
2) Linesearch should return an error if it cannot find a suitable step size (i.e., one that at least decreases function value), it indicates that the gradient tolerance is too stringent or that the linesearch does not handle floating point well. If it can, then MajorIteration is indicated and that found location should be checked for convergence.

1) Can Bisection's comparison work in the absence of gradient? I am a bit hesitant about introducing functions that deal with floating point effect into a high-level function like minimize(). Also, I feel that it is hard to come up with a test without knowing how the location were generated.
2) At the moment I would keep the simple test `if loc.F <= optLoc.F`. When we have Nelder-Mead, that behaves very differently from line-search methods, we can again revisit this. Right now I would like to fix and clean up Local() and minimize() so that the tests in the other PR pass.

vladimir-ch:
An additional comment about Stats.GradientNorm: I wouldn't move it Location after all. It is the maximum norm that is being used only in the convergence check. If it were a 2-norm, then having it in Location would be useful also in the minimizers (e.g., in CG) and I would prefer to have it there.

As Stats are now, convergence check is based solely on its data fields, which is nice. I would even lean towards making checkConvergence() a method on Stats, then Stats could have a non-exported field to keep the function value history for FunctionRelTol test.

Morever, I think it would be best to update stats.GradientNorm just before checking gradient convergence, in this way it would be guaranteed to be the value that triggers the convergence and its manipulation would be kept to a single place.

btracey:
> On Jan 18, 2015, at 5:00 PM, Vladimír Chalupecký <notifications@github.com> wrote:
> 
> An additional comment about Stats.GradientNorm: I wouldn't move it Location after all. It is the maximum norm that is being used only in the convergence check. If it were a 2-norm, then having it in Location would be useful also in the minimizers (e.g., in CG) and I would prefer to have it there.
> 
SGTM
> As Stats are now, convergence check is based solely on its data fields, which is nice. I would even lean towards making checkConvergence() a method on Stats, then Stats could have a non-exported field to keep the function value history for FunctionRelTol test.
> 
As we’ve discussed though, the gradient check should really be on OptLoc, unless we’re moving OptLoc into stats.

We may need some kind of interface for FunctionRelTol like we do for Recorder. There are so many different ways to implement that kind of convergence. The most common would be that you haven’t seen an improvement of eps in Y evaluations, but there are many such flavors. I would imagine there is a big overlap between this and a future Global optimization function, where the only reasonable global convergence metrics have to do with the number of function evaluations and the last time that significant improvement has been seen. It’s worth thinking about that when doing the design.

vladimir-ch:
Acceptable changes from this PR have been submitted, reviewed and merged separately, so closing.

PR #45, opened 2015-01-20
Handle NoEvaluation in evaluate()

vladimir-ch:
Currently, evaluate() panics when `evalType == NoEvaluation`. Also, location.X is modified even when evalType is not supported and no evaluation takes place. This PR fixes both issues and improves the comment for evaluate().

### DIFF
--- a/local.go
+++ b/local.go
@@ -5,6 +5,7 @@
 package optimize
 
 import (
+	"fmt"
 	"math"
 	"time"
 
@@ -136,11 +137,7 @@ func minimize(settings *Settings, location *Location, method Method, funcInfo *F
 		}
 
 		// Compute the new function and update the statistics
-		err = evaluate(funcs, funcInfo, evalType, xNext, location, stats)
-		if err != nil {
-			status = Failure
-			return
-		}
+		evaluate(funcs, funcInfo, evalType, xNext, location, stats)
 		update(location, optLoc, stats, iterType, startTime)
 
 		// Find the next location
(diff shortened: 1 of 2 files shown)


local.go
```diff
@@ -275,8 +275,17 @@ func checkConvergence(loc *Location, itertype IterationType, stats *Stats, setti
 	return NotTerminated
 }
 
-// evaluate evaluates the function and stores the answer in place
+// evaluate evaluates the function given by funcs at xNext, stores the answer
+// in location and updates stats. If the function does not support the
```
btracey: s/in/into

local.go
```diff
@@ -275,8 +275,17 @@ func checkConvergence(loc *Location, itertype IterationType, stats *Stats, setti
...
+// requested evalType, location and stats are not modified and ErrMismatch is
+// returned.
 func evaluate(funcs functions, funcInfo *FunctionInfo, evalType EvaluationType, xNext []float64, location *Location, stats *Stats) error {
+	if evalType == NoEvaluation {
+		return nil
+	}
+	if !funcInfo.Supports(evalType) {
```
btracey: I think this should panic. If the user lets us choose an optimizer, this will never err (unless we have a bug). If the user chooses an optimizer, then it's a mismatch in parameters that is verifiable before the optimization is used. It's also not quite the right way to do it in the general case. For example, if the function is constrained and the user chooses Lbfgs, there should be an err/panic. In the end, it's not about Function supporting EvalType, but Method supporting the specific Function. 

Given we do use some magic, I can see the need for support. It seems like the better case is for the optimizer to have a Support function that either returns a boolean (if it supports), or an error which is nil if it supports, and provides a detailed explanation if it isn't, i.e. "Bfgs requires the Function to have a Gradient. Function does not implement either Gradient or FunctionGradient."

local.go
```diff
@@ -275,8 +275,17 @@ func checkConvergence(loc *Location, itertype IterationType, stats *Stats, setti
 	return NotTerminated
 }
 
-// evaluate evaluates the function and stores the answer in place
+// evaluate evaluates the function given by funcs at xNext, stores the answer
+// in location and updates stats. If the function does not support the
+// requested evalType, location and stats are not modified and ErrMismatch is
+// returned.
 func evaluate(funcs functions, funcInfo *FunctionInfo, evalType EvaluationType, xNext []float64, location *Location, stats *Stats) error {
+	if evalType == NoEvaluation {
```
btracey: This should do nothing if sameX, and nan out if not, right?

local.go
vladimir-ch: > I think this should panic.

Agreed. I just used the error that had already been there.

> For example, if the function is constrained and the user chooses Lbfgs, there should be an err/panic.

If the function is also a Function, then it does not necessarily need to err/panic in that case. This is something to keep for later.

> In the end, it's not about Function supporting EvalType, but Method supporting the specific Function.

Agreed. I will remove Supports() from FunctionInfo.

> Given we do use some magic, I can see the need for support. It seems like the better case is for the optimizer to have a Support function that either returns a boolean (if it supports), or an error which is nil if it supports, and provides a detailed explanation if it isn't, i.e. "Bfgs requires the Function to have a Gradient. Function does not implement either Gradient or FunctionGradient."

Good idea, I will put it into Issues as an enhancement.

local.go
vladimir-ch: Yea, I was not sure about the right behavior in this case. I wanted NoEvaluation to be a no-op, but your suggestion is also quite (even more) reasonable: who wants a no-op can choose not call evaluate() when NoEvaluation. I will modify accordingly.

vladimir-ch:
Updated. PTAL, @btracey

local.go
```diff
@@ -276,32 +277,41 @@ func checkConvergence(loc *Location, itertype IterationType, stats *Stats, setti
 	return NotTerminated
 }
 
-// evaluate evaluates the function and stores the answer in place
+// evaluate evaluates the function given by funcs at xNext, stores the answer
+// into location and updates stats. If location.X is not equal to xNext and
+// evalType == NoEvaluation, location is filled with NaN values. If the
```
btracey: Things are filled with NaN values for lots of reasons. How about 

"If location.X is not equal to xNext, then unused fields of Location are set to NaN"

local.go
```diff
@@ -276,32 +277,41 @@ func checkConvergence(loc *Location, itertype IterationType, stats *Stats, setti
...
+// function does not support the requested evalType, location and stats are not
```
btracey: This comment is out of date. It now panics.

local.go
```diff
@@ -321,8 +337,22 @@ func evaluate(funcs functions, funcInfo *FunctionInfo, evalType EvaluationType,
...
+				location.X[i] = math.NaN()
+			}
+			for i := range location.Gradient {
+				location.Gradient[i] = math.NaN()
+			}
+		}
+		return nil
 	default:
-		panic("unreachable")
+		panic(fmt.Sprintf("unknown evaluation type %v", int(evalType)))
```
btracey: maybe just "unknown evaluation type " + evalType.String()

local.go
vladimir-ch: Well, it does not panic yet, only when the evaluation type is unknown. If the evaluation type does not match the function, it returns an error, but I will change it to a panic.

local.go
vladimir-ch: Yeah, I wanted to do this, but it means modifying EvaluationType.String(), because ATM it would panic for unknown evaluation types. I will modify it.

local.go

local.go

local.go

vladimir-ch:
PTAL, @btracey

types.go
```diff
@@ -21,6 +22,9 @@ const (
 )
 
 func (e EvaluationType) String() string {
+	if e < 0 || int(e) >= len(evaluationStrings) {
+		return fmt.Sprintf("EvaluationType(%d)", e)
```
btracey: Does this work, or does this need to be int(e)?

btracey:
LGTM when last comment is addressed.

types.go
vladimir-ch: Yes, it works, because 0 is an untyped integer constant that can be compared or assigned to any compatible integer variable.

vladimir-ch: "compatible" is superfluous, just any integer variable.

btracey: Sorry,  that wasn't my question. I am asking about the fmt call. e is a
Stringer. I think to print the actual integer value you need to cast to
int. Maybe that is overwritten by %d

vladimir-ch: Oh, sorry. Anyway, it works, but I can add an explicit conversion.

kortschak: Only verbs 'v', 's', 'x', 'X', 'q' will call the String(). 'd' will not. Conversion is not necessary.

PR #47, opened 2015-01-20
Rewrite Local()

vladimir-ch:
The first step towards fixing `Local()`, `minimize()` and `Linesearch`. I tried to keep the changes limited to `Local()`, but it was not possible everywhere.

Summary of changes:

* Put various initializations into a more reasonable order, so that things are initialized only when necessary.
* Properly record the starting location.
* Check the starting location for convergence before calling `minimize()`
* Check the initial function status if it is a `Statuser`
* Reduce the number of arguments to `minimize()`
* `getStartingLocation()` now also returns `EvaluationType` that was used at `initX`
* Add implementation comments to `Local()`

### DIFF
--- a/local.go
+++ b/local.go
@@ -54,59 +54,79 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
 	}
 
 	startTime := time.Now()
-
 	funcs, funcInfo := getFunctionInfo(f)
 
-	if method == nil {
-		method = getDefaultMethod(funcInfo)
+	if funcInfo.IsStatuser {
+		_, err := funcs.status.Status()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if settings.Recorder != nil {
+		// Initialize Recorder first. If it fails, we avoid the (possibly
+		// time-consuming) evaluation of F and DF at the starting location.
+		err := settings.Recorder.Init(funcInfo)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	if settings == nil {
 		settings = DefaultSettings()
 	}
-
 	stats := &Stats{}
-	location, err := getStartingLocation(funcs, funcInfo, initX, stats, settings)
+	optLoc, evalType, err := getStartingLocation(funcs, funcInfo, initX, stats, settings)
 	if err != nil {
 		return nil, err
 	}
 
-	optLoc := &Location{}
-	// update stats (grad norm, function value, etc.) so that things are
-	// initialized for the first convergence check
-	update(location, optLoc, stats, NoIteration, startTime)
-
+	// Runtime is the only Stats field that needs to be updated here.
+	stats.Runtime = time.Since(startTime)
+	// Send optLoc to Recorder before checking it for convergence.
 	if settings.Recorder != nil {
-		err = settings.Recorder.Init(funcInfo)
-		if err != nil {
-			return &Result{Status: Failure}, err
-		}
+		// TODO(vladimir-ch): Replace NoIteration with InitialIteration when it
+		// is added.
+		err = settings.Recorder.Record(optLoc, evalType, NoIteration, stats)
 	}
 
-	// actually perform the minimization
-	status, err := minimize(settings, location, method, funcInfo, stats, funcs, optLoc, startTime)
+	// Check if the starting location satisfies the convergence criteria.
+	// TODO(vladimir-ch): Replace MajorIteration with InitialIteration when it
+	// is added and checkConvergence() updated to handle it.
+	status := checkConvergence(optLoc, MajorIteration, stats, settings)
+	if status == NotTerminated && err == nil {
+		if method == nil {
+			method = getDefaultMethod(funcInfo)
+		}
+		// The starting location is not good enough, we need to perform a
+		// minimization. The optimal location will be stored in-place in
+		// optLoc.
+		status, err = minimize(settings, method, funcInfo, stats, funcs, optLoc, startTime)
+	}
 
-	// cleanup at exit
 	if settings.Recorder != nil && err == nil {
+		// Send the optimal location to Recorder.
 		err = settings.Recorder.Record(optLoc, NoEvaluation, PostIteration, stats)
 	}
 	stats.Runtime = time.Since(startTime)
 	return &Result{
-		Stats:    *stats,
 		Location: *optLoc,
+		Stats:    *stats,
 		Status:   status,
 	}, err
 }
 
-func minimize(settings *Settings, location *Location, method Method, funcInfo *FunctionInfo, stats *Stats, funcs functions, optLoc *Location, startTime time.Time) (status Status, err error) {
-	methodStatus, methodIsStatuser := method.(Statuser)
+func minimize(settings *Settings, method Method, funcInfo *FunctionInfo, stats *Stats, funcs functions, optLoc *Location, startTime time.Time) (status Status, err error) {
+	location := &Location{}
+	copyLocation(location, optLoc)
 	xNext := make([]float64, len(location.X))
+
+	methodStatus, methodIsStatuser := method.(Statuser)
 
 	evalType, iterType, err := method.Init(location, funcInfo, xNext)
 	if err != nil {
 		return Failure, err
 	}
-	copyLocation(optLoc, location)
 
 	for {
 		if settings.Recorder != nil {
@@ -191,7 +211,7 @@ func getDefaultMethod(funcInfo *FunctionInfo) Method {
 }
 
 // getStartingLocation allocates and initializes the starting location for the minimization.
-func getStartingLocation(funcs functions, funcInfo *FunctionInfo, initX []float64, stats *Stats, settings *Settings) (*Location, error) {
+func getStartingLocation(funcs functions, funcInfo *FunctionInfo, initX []float64, stats *Stats, settings *Settings) (*Location, EvaluationType, error) {
 	dim := len(initX)
 	l := &Location{
 		X: make([]float64, dim),
@@ -201,6 +221,7 @@ func getStartingLocation(funcs functions, funcInfo *FunctionInfo, initX []float6
 	}
 	copy(l.X, initX)
 
+	evalType := NoEvaluation
 	if settings.UseInitialData {
 		l.F = settings.InitialFunctionValue
 		if l.Gradient != nil {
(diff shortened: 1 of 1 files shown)


vladimir-ch:
Could you PTAL, @btracey ?

local.go
```diff
@@ -54,59 +54,92 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
-	// update stats (grad norm, function value, etc.) so that things are
-	// initialized for the first convergence check
-	update(location, optLoc, stats, NoIteration, startTime)
-
+	// Send optLoc to Recorder before checking it for convergence.
 	if settings.Recorder != nil {
-		err = settings.Recorder.Init(funcInfo)
+		// Runtime is the only Stats field that needs to be updated here.
+		stats.Runtime = time.Since(startTime)
+		// TODO: Replace NoIteration with InitialIterate when it is added.
```
btracey: InitialIteration

local.go
```diff
@@ -54,59 +54,92 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
-	optLoc := &Location{}
-	// update stats (grad norm, function value, etc.) so that things are
-	// initialized for the first convergence check
-	update(location, optLoc, stats, NoIteration, startTime)
-
+	// Send optLoc to Recorder before checking it for convergence.
 	if settings.Recorder != nil {
-		err = settings.Recorder.Init(funcInfo)
+		// Runtime is the only Stats field that needs to be updated here.
+		stats.Runtime = time.Since(startTime)
```
btracey: Runtime can be before the if, as it's called both in the if and out of it.

local.go
```diff
@@ -54,59 +54,92 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
+			// Build a meaningful result with Failure status and return it.
+			return &Result{*optLoc, *stats, Failure}, err
 		}
 	}
 
-	// actually perform the minimization
-	status, err := minimize(settings, location, method, funcInfo, stats, funcs, optLoc, startTime)
+	// Check if the starting location satisfies the convergence criteria.
+	stats.Runtime = time.Since(startTime)
+	// TODO: Replace MajorIteration with InitialIteration when it is added and
```
btracey: TODO should have your handle with it.

TODO(vladimir-ch):

local.go
```diff
@@ -54,59 +54,92 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
+	status := checkConvergence(optLoc, MajorIteration, stats, settings)
+	if status == NotTerminated {
+		// The starting location is not good enough, we need to start the
+		// minimization run.
+
+		// First, check the status of the function to see if it is able to
+		// evaluate.
+		// TODO: Is this the best place for checking the function status?
+		if funcInfo.IsStatuser {
+			// TODO: should the function really return status?
```
btracey: I think so. It allows the function to signal completion for some reason. The only other way to communicate that is by falsely returning 0 for the gradient (or something). Two good reasons to have this:
1) The function can have an error for some reason. This is a way to put in that check without having to do it at every function evaluation where there is nothing the optimizer can do about it
2) In machine learning, one often wants to converge until the testing error is minimized (not the training error). The machine learning parameter function can return a status signifying early convergence even though a local minima has not been found.

local.go
```diff
@@ -54,59 +54,92 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
+			status, err = funcs.status.Status()
+			if err != nil || status != NotTerminated {
+				// Function status is bad, but at least we have the starting
```
btracey: I think this may be a good case for a goto. Have line 119 be a "Cleanup:" label and jump to it in this f. The recorder still can get useful information here if err != nil && status != NotTerminated.

btracey: Status doesn't have to be bad. "// Function status says optimization should be terminated"

local.go
btracey: (This applies globally)

local.go
```diff
@@ -346,6 +382,7 @@ func update(location *Location, optLoc *Location, stats *Stats, iterType Iterati
 	if iterType == MajorIteration {
 		stats.MajorIterations++
 	}
+	// TODO: Update optLoc only at MajorIterations?
 	if location.F < optLoc.F {
```
btracey: This should be <=

local.go
```diff
@@ -138,6 +171,7 @@ func minimize(settings *Settings, location *Location, method Method, funcInfo *F
 
 		// Compute the new function and update the statistics
 		evaluate(funcs, funcInfo, evalType, xNext, location, stats)
+		// TODO: Check for NaN.
```
btracey: I don't think NaN/Inf is necessarily a problem. Getting it on major iterations is bad, but the linesearchers should be smart enough to backtrack form that (and if they aren't we should make them as such). NaN and Inf are a problem at the starting location because gradients don't make any sense.

local.go
vladimir-ch: or InitIteration?

local.go
vladimir-ch: Changed

local.go
vladimir-ch: OK

local.go
vladimir-ch: OK, removed the line.

local.go
vladimir-ch: goto and label Cleanup added, comment updated.

local.go

local.go
vladimir-ch: OK, I will add an issue to make sure linesearchers handle NaN well.

vladimir-ch:
I addressed the comments.

I added one check at the beginning of Local(): if f is a Statuser and Status() returns an error, panic.

PTAL, @btracey

local.go
```diff
@@ -54,59 +54,100 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
-
 	funcs, funcInfo := getFunctionInfo(f)
+	status := NotTerminated
+
+	if funcInfo.IsStatuser {
+		_, err := funcs.status.Status()
+		if err != nil {
+			// Function status is bad, panic.
+			// TODO: Document this panic in the function comment.
+			panic(fmt.Sprintf("bad initial function status: %v", err))
```
btracey: This doesn't feel to me like it should panic. Since an error value is returned , it can presumably be handled, which means it should be returned.

btracey: Honestly though, I'm okay with not having this call at all. Right now the flow is simple: 1) Call the function to generate the initial location if necessary. 2) Check the status before each step of the minimization, returning the Status and error. Status is checked before the first step of the minimization, so all this does is catch an error if the initial condition is not specified. If I were a function writer, I would have the returned function value be NaN or Inf if the function was actually in error, both of which will create an error if no initial condition is specified. If the initial condition is specified, we already check Status before the first step and do the reasonable thing.

If you do want this check here, it should behave the same as all the other checks (i.e. go to Cleanup).

btracey: "return nil, err" like in the Recorder check is better.

local.go
```diff
@@ -54,59 +54,100 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
+	status = checkConvergence(optLoc, MajorIteration, stats, settings)
+	if status == NotTerminated {
+		// The starting location is not good enough, we need to start the
+		// minimization run.
+
+		// First, check the status of the function to see if it is able to
+		// evaluate.
+		// TODO(vladimir-ch): Is this the best place for checking the function
+		// status? Revise after minimize() is updated.
+		if funcInfo.IsStatuser {
```
btracey: Is this check actually necessary? It already gets checked before the next function evaluation in minimize. Removing it would make this much simpler and would also allow the removal of the Cleanup label (sorry for not seeing this before).

local.go
```diff
@@ -54,59 +54,100 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
...
 	}
 
-	// actually perform the minimization
-	status, err := minimize(settings, location, method, funcInfo, stats, funcs, optLoc, startTime)
+	// Check if the starting location satisfies the convergence criteria.
+	// TODO(vladimir-ch): Replace MajorIteration with InitialIteration when it
+	// is added and checkConvergence() updated to handle it.
+	status = checkConvergence(optLoc, MajorIteration, stats, settings)
+	if status == NotTerminated {
+		// The starting location is not good enough, we need to start the
```
btracey: need to perform a minimization.

local.go
```diff
@@ -346,7 +389,8 @@ func update(location *Location, optLoc *Location, stats *Stats, iterType Iterati
 	if iterType == MajorIteration {
 		stats.MajorIterations++
 	}
-	if location.F < optLoc.F {
+	// TODO: Update optLoc only at MajorIterations?
```
btracey: I don't think we want to do this. For example, a linesearch may find a better function value but not have sufficient gradient decrease to meet Armijio. It wants to continue the linesearch, but the function evaluation limit is reached. We want to return the best point found in this case I would imagine.

local.go
vladimir-ch: The motivation for this check was to make sure that the function has a good status before calling getStartingLocation(). The motivation for the panic was that if we return an error, the caller should be able to handle it, so if he is able to handle it after, he can also handle it before and should not call Local() with bad function status in the first place. I do not insist on this, I put in there as a possibility and to discuss it.

I will change it to "return nil, err".

local.go
vladimir-ch: SGTM

local.go
vladimir-ch: Changed.

local.go
vladimir-ch: No, it is not necessary. I wanted to revisit this after reordering the calls in minimize(), but it is already clear that funcInfo.status will be checked before calling evaluate(). Removed.

local.go
vladimir-ch: Changed.

local.go
```diff
@@ -260,6 +281,7 @@ func checkConvergence(loc *Location, itertype IterationType, stats *Stats, setti
 	}
 
 	if settings.Runtime > 0 {
+		// TODO: It would be nice to update Runtime here.
```
btracey: This also needs a (vladimir-ch)

btracey:
LGTM with the last comment addressed.

PR #50, opened 2015-01-23
Reorganize the main for loop in minimize()

vladimir-ch:


local.go
```diff
@@ -129,38 +129,42 @@ func minimize(settings *Settings, method Method, funcInfo *FunctionInfo, stats *
...
+		evaluate(funcs, funcInfo, evalType, xNext, location, stats)
+		// Update the stats and optLoc (provided a new optimum has been found).
+		update(location, optLoc, stats, iterType, startTime)
+		// Get the convergence status before recording the new location.
+		status = checkConvergence(optLoc, iterType, stats, settings)
+
 		if settings.Recorder != nil {
 			err = settings.Recorder.Record(location, evalType, iterType, stats)
 			if err != nil {
-				status = Failure
```
btracey: Why wouldn't we set the Status?

vladimir-ch: I want to use the status returned from checkConvergence(), not overwrite it with Failure.

btracey: I guess that's reasonable. I was hesitant because I was pondering the implications of seeing NotTerminated as a status externally, but I suppose it's fine.

btracey:
LGTM

local.go
vladimir-ch: I think there are not yet any implications, so we can still freely decide what is being returned. Getting NotTerminated and an error seems very natural to me.

PR #51, opened 2015-01-23
Return correct iteration types from Linesearch

vladimir-ch:
* Fixed a missing return in `evaluate()`
* `finishedF` is not needed, because `evaluate()` will not overwrite our location
* Removed `finished` and replaced with `nextIterType`.
* Those copy() on non-MinorIterations are not strictly necessary.

The current `go test` passes, but interestingly its runtime has increased slightly, from roughly 0.4s on my machine to 0.46s. I have not investigated the reason.

The updated tests in the other PR pass too.

### DIFF
--- a/linesearch.go
+++ b/linesearch.go
@@ -24,8 +24,7 @@ type Linesearch struct {
 	funInfo *FunctionInfo
 
 	lastEvalType EvaluationType
-	finished     bool
-	finishedF    float64
+	iterType     IterationType
 }
 
 func (l *Linesearch) Init(loc *Location, f *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
@@ -50,15 +49,22 @@ func (l *Linesearch) Init(loc *Location, f *FunctionInfo, xNext []float64) (Eval
 	floats.AddScaledTo(xNext, l.initLoc, stepSize, l.direction)
 	l.funInfo = f
 	l.lastEvalType = evalType
-	l.finished = false
-	return evalType, MajorIteration, nil
+	l.iterType = MinorIteration
+	return evalType, l.iterType, nil
 }
 
 func (l *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
-	if l.finished {
-		// Means that we needed to evaluate the gradient, so now we have it and can initialize
-		l.finished = false
-		loc.F = l.finishedF
+	if l.iterType == SubIteration {
+		// We needed to evaluate the gradient, so now we have it and can
+		// announce MajorIteration.
+		l.iterType = MajorIteration
+		copy(xNext, loc.X)
+		return NoEvaluation, l.iterType, nil
+	}
+	if l.iterType == MajorIteration {
+		// The linesearch previously signaled MajorIteration. Since we're here,
+		// it means that the previous location is not good enough to converge,
+		// so start the next linesearch.
 		return l.initNextLinesearch(loc, xNext)
 	}
 	projGrad := math.NaN()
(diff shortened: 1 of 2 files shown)


vladimir-ch:
I think that those copy() on non-MinorIterations are not necessary as the code is now.

linesearch.go
```diff
@@ -50,15 +49,20 @@ func (l *Linesearch) Init(loc *Location, f *FunctionInfo, xNext []float64) (Eval
...
 
 func (l *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
-	if l.finished {
-		// Means that we needed to evaluate the gradient, so now we have it and can initialize
-		l.finished = false
-		loc.F = l.finishedF
+	if l.nextIterType == SubIteration {
+		// We needed to evaluate the gradient, so now we have it and can
+		// announce MajorIteration.
+		l.nextIterType = MajorIteration
```
btracey: While true for conjugate gradient, this is wrong in the general case. For example, if the gradient is very expensive to evaluate, a linesearch may first evaluate the function. If the function value is worse, it knows it should backtrack, otherwise it then evaluates the gradient to see if it meets the armijio condition. Just doing a two-stage evaluation of the gradient does not signify a MajorIteration -- the linesearch converging is the only signifier.

btracey: It worked before because we had explicitly checked if the linesearch had finished.

linesearch.go
```diff
@@ -50,15 +49,20 @@ func (l *Linesearch) Init(loc *Location, f *FunctionInfo, xNext []float64) (Eval
...
+		copy(xNext, loc.X)
+		return NoEvaluation, l.nextIterType, nil
+	}
+	if l.nextIterType == MajorIteration {
+		// No convergence at MajorIteration, so continue iterating.
```
btracey: I would prefer something along the lines of

"// The linesearch previously signaled MajorIteration. Since we're here, it means that the previous location is not good enough to converge, so start the next linesearch."

linesearch.go
```diff
@@ -71,29 +75,30 @@ func (l *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, It
...
 			// We have the function value at the current location, but we don't
-			// have the gradient, so get it before starting the next major iteration.
+			// have the gradient, so get it before announcing MajorIteration.
+			l.nextIterType = SubIteration
 			copy(xNext, loc.X)
-			return GradientEval, SubIteration, nil
+			return GradientEval, l.nextIterType, nil
 		}
-		return l.initNextLinesearch(loc, xNext)
+		// No gradient evaluation is necessary, announce MajorIteration.
```
btracey: // The linesearch is finished. Announce so with an update to MajorIteration. The function value and gradient is already known, so no function evaluations are necessary.

linesearch.go
```diff
@@ -71,29 +75,30 @@ func (l *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, It
...
-	// Line search not done, just iterate
+	// Line search not done, just iterate.
 	stepSize, evalType, err := l.Method.Iterate(linesearchLocation)
 	if err != nil {
 		return NoEvaluation, NoIteration, err
 	}
 	floats.AddScaledTo(xNext, l.initLoc, stepSize, l.direction)
 	l.lastEvalType = evalType
-	return evalType, MinorIteration, nil
+	return evalType, l.nextIterType, nil
```
btracey: I know it's not strictly necessary, but could you have a 
````
l.nextIterType = MinorIteration
````
right after the l.lastEvalType line? It makes it more clear than having to remember it was set in initNextLinesearch and then trying to ensure it didn't get changed somehow.

File: linesearch.go (hunk 1 above)
Comment (btracey) [inline linesearch.go]: Sorry, you're right, this works. Linesearch is in control of the IterationType, so it is the only producer of SubIteration.

File: linesearch.go [hunk 5]
```diff
@@ -71,29 +75,30 @@ func (l *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, It
 	}
 	if l.Method.Finished(linesearchLocation) {
 		if l.lastEvalType == FunctionEval {
-			l.finished = true
```
btracey: I think the previous check was wrong. This should be 
````
if l.lastEvalType == FunctionEval && loc.Gradient != nil{
}
````
Right? If it's a gradient-free function then we don't need to evaluate the gradient.

linesearch.go
vladimir-ch: Changed.

linesearch.go
vladimir-ch: Changed.

linesearch.go
vladimir-ch: Added.

linesearch.go
vladimir-ch: Yes, that's right, good catch.

vladimir-ch:
PTAL, @btracey .

Is it ok if I change nextIterType to just iterType?

btracey:
Yes

btracey:
LGTM

PR #55, opened 2015-01-26
Fix comment formatting

vladimir-ch:


vladimir-ch:
Merging directly because the fix is trivial.

btracey:
I know it's trivial, but could you wait for a LGTM in the future? "All changes need a LGTM" is a much easier guideline that "changes need a LGTM unless they're trivial and you know they don't break things".

Issue #53, opened 2015-01-23
Remove cleverness in Bisection linesearch


btracey:
The function comparison trickiness in Bisection causes difficulty with the operation of the rest of the package. Furthermore, it doesn't even work in the general case -- the floating point noise depends on the specific function. It's just not possible to converge a general non-linear function that far. Instead, we should at least describe (and possibly provide) a simple way to converge the gradient past 1e-10 if the user is willing to assume the function is locally quasi-convex.

PR #57, opened 2015-01-26
Remove extra check for conclusion of Bisection.

btracey:
The original code for Bisection included a secondary check to the strong Wolfe conditions to see if the optmiization was finished. The idea was to help mitigate floating point noise and allow for stronger convergence to the gradient. Unfortunately, all this does is add complexity. The parameters are ad hoc, and trade off floating point noise for actual function modulation. For more complicated functions (especially concurrent ones) the noise will be higher, while other functions may have modulations that are very small. It is impossible to design a tradeoff that is good for all functions. Instead, keep the code simple. This also fixes issues with Bisection and the outer OptLoc disagreeing on the optimum location

### DIFF
--- a/bisection.go
+++ b/bisection.go
@@ -4,11 +4,7 @@
 
 package optimize
 
-import (
-	"math"
-
-	"github.com/gonum/floats"
-)
+import "math"
 
 // Bisection is a LinesearchMethod that uses a bisection to find a point that
 // satisfies the strong Wolfe conditions with the given gradient constant and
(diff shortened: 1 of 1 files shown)


btracey:
This will fix #53 . PTAL @vladimir-ch

bisection.go
```diff
@@ -60,18 +56,7 @@ func (b *Bisection) Init(initLoc LinesearchLocation, initStepSize float64, f *Fu
...
-)
-
 func (b *Bisection) Finished(l LinesearchLocation) bool {
-	if floats.EqualWithinRel(l.F, b.initF, funcSmallEqual) && math.Abs(l.Derivative) < gradSmallEqual && math.Abs(b.initGrad) < gradSmallEqual {
-		// The two numbers are so close that we should just check on the gradient
-		// TODO: Should iterate be updated? Maybe find a function where it needs it.
-		return math.Abs(l.Derivative) < b.GradConst*math.Abs(b.initGrad)
-	}
-
 	return StrongWolfeConditionsMet(l.F, l.Derivative, b.initF, b.initGrad, b.currStep, 0, b.GradConst)
```
vladimir-ch: Not related to the proposed changes, but why is the Armijo constant equal to 0?

btracey: Strictly, the strong Wolfe conditions only hold if FunConst > 0. However, it can be any number greater than zero, including very small values. In finite precision world, when we get close to the minimum, there are large regions where the function value is "constant" as far as representable float64 are concerned. In this region, the line search has to fail because there are no locations with a smaller function value, even though with infinite precision there would be many steps that satisfy the condition. Setting C1 to zero avoids those problems. On a conceptual level, we care about the function not increasing and the gradient getting smaller. Since these methods assume smooth and continuous functions, driving the gradient to zero is the same as finding the location with minimum F. We're happy stopping the linesearch with a sufficient decrease in gradient norm.

vladimir-ch: OK, thanks.

vladimir-ch:
LGTM.

You may already know that, but you can close the issue #53 by modifying the commit message. It is described in https://help.github.com/articles/closing-issues-via-commit-messages/ It is quite useful and leaves an explicit reference in the commit message to this PR. Basically, you would have to do something like this:

```
git checkout fixbisectio
git commit --amend
(add "Closes #53" at the end of the commit message and save)
git push -f (push to github so that it recognizes when the branch has been merged)
git checkout master
git merge fixbisectio
git push
(issue #53 and this PR are closed automatically)
```

PR #59, opened 2015-01-26
Fix order of DefaultSettings

btracey:
The recorder is checked before the settings are allocated, so this causes a nil pointer dereference if settings == nil


btracey:
PTAL @vladimir-ch

vladimir-ch:
The change in local.go LGTM, but you added by mistake also neldermead.go.

PR #62, opened 2015-01-30
Unify variable names

vladimir-ch:
* `Location` variables were named `location`, `loc` and `l` at different places in optimize, which was confusing. So this PR unifies `Location` variables to be consistently `loc`.
* `LinesearchLocation` variables are renamed to `loc`, `lsLoc` or `initLoc`, depending on the context. 
* `Linesearch` receiver is renamed to `ls`, `l` seemed too short, especially when `Location`s were still `l` at some places.
* Redundant `[]float64` from `Df()` and `FDf()` are removed.

This PR is simple, but extensive, so it is better reviewed commit by commit.

Remaining inconsistencies:

* `direction` and `dir` (let's use `dir`)
* `initLoc`, `initX`, `currStep` ... vs. `xNext`, `fPrev`, `stepSizePrev`, ... (let's use the abbreviated adjective at the beginning, i.e., `nextX`, `prevF`, ...)

### DIFF
--- a/backtracking.go
+++ b/backtracking.go
@@ -32,11 +32,11 @@ type Backtracking struct {
 	initG    float64
 }
 
-func (b *Backtracking) Init(initLoc LinesearchLocation, initStepSize float64, f *FunctionInfo) EvaluationType {
-	if initStepSize <= 0 {
+func (b *Backtracking) Init(loc LinesearchLocation, step float64, _ *FunctionInfo) EvaluationType {
+	if step <= 0 {
 		panic("backtracking: bad step size")
 	}
-	if initLoc.Derivative >= 0 {
+	if loc.Derivative >= 0 {
 		panic("Backtracking: init G non-negative")
 	}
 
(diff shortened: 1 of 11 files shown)


kortschak:
There are a number of places where the `LineSearchLocation`/`*Location` is not used in the func. It would be informative if the var names in those places were `_`.

What is the reason for naming the `LinesearchLocation` `initiLoc` in the interface definition?

btracey:
I'm with Dan, I think "loc" is better in the interface definition. It's loc in the other Init functions.

vladimir-ch:
I will change it.

Additional questions from looking at interfaces.go:
* `LinesearchMethod.Init()` takes `step`, `NextDirectioner` returns `stepSize`. They should have the same name, preferrably stepSize, because of `StepSizer.StepSize()`. Agreed?
* `Recorder`'s methods have unnamed arguments. Does it matter or should they have names?
* What about the adjectives-before-or-after issue? In interfaces.go there is `nextStep`, but also `xNext`. Should it be unified or it doesn't matter?

btracey:
- If we're picking one, I think StepSize is preferable to Step. Reason: Step can be a verb, whereas StepSize is clearly a noun. I'd also be okay with all of the variable names just being step. Given that it's a float64, it's clear that it's a noun and not an adjective.
- My take on Go style has been that named arguments are good if they help. I don't think named arguments add any clarity for Recorder. 
- xNext is a personal convention I have used for a long time, so it looks more natural to me. The reason for the convention is that previous codes have had several small items that need to be interated, and nextX and nextY (for example) look too similar to me. Having the small thing first at least highlights the difference between xNext and yNext. With the longer ones, nextStep flows more naturally in english, and because the second word is longer it's easier to distinguish. That's my argument anyway, may not be a good one for the new era.

vladimir-ch:
> If we're picking one, I think StepSize is preferable to Step. Reason: Step can be a verb, whereas StepSize is clearly a noun. I'd also be okay with all of the variable names just being step. Given that it's a float64, it's clear that it's a noun and not an adjective.

I am also ok with just `step`, its short and nice. Also, adding `next` or `prev` to `stepSize` can make it too long. So, `step`?

> My take on Go style has been that named arguments are good if they help. I don't think named arguments add any clarity for Recorder.

That's what I thought. I'll leave Recorder as it is.

> xNext is a personal convention I have used for a long time, so it looks more natural to me...

I've got used to it too, I was just asking. There is also `initX` that should be xInit according to this argument, but I guess this inconsistence is not that important.

vladimir-ch:
Summary of the recent commits:

* Use `dir` consistently instead of `direction` for the descent direction vector.
* Remove unused field `direction` in BFGS.
* Rename unused arguments to _
* Use `step` instead of `stepSize` in interfaces.go

PTAL @btracey @kortschak

kortschak:
LGTM but wait for @btracey.

btracey:
LGTM

PR #63, opened 2015-02-02
Add testopt, a testing subpackage

vladimir-ch:
* Created testopt subpackage
* Copied objective functions from unconstrained_test.go
* Improved their documentation
* Most importantly, added tests that revealed mistakes in four functions. http://people.sc.fsu.edu/~jburkardt/m_src/test_opt/test_opt.html was used as a reference implementation
* Functions now return a list of their known minima

Modifying unconstrained_test.go to actually use these new, fairly reliably correct functions will come in another PR.

### DIFF
--- a/functions/functions.go
+++ b/functions/functions.go
@@ -1,0 +1,1448 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package functions
+
+import (
+	"math"
+
+	"github.com/gonum/floats"
+)
+
+// Beale implements the Beale's function.
+//
+// Standard starting points:
+//  Easy: [1, 1]
+//  Hard: [1, 4]
+//
+// References:
+//  - Beale, E.: On an Iterative Method for Finding a Local Minimum of a
+//    Function of More than One Variable. Technical Report 25, Statistical
+//    Techniques Research Group, Princeton University (1958)
+//  - More, J., Garbow, B.S., Hillstrom, K.E.: Testing unconstrained
+//    optimization software. ACM Trans Math Softw 7 (1981), 17-41
+type Beale struct{}
+
+func (Beale) F(x []float64) float64 {
+	if len(x) != 2 {
+		panic("dimension of the problem must be 2")
+	}
+
+	f1 := 1.5 - x[0]*(1-x[1])
+	f2 := 2.25 - x[0]*(1-x[1]*x[1])
+	f3 := 2.625 - x[0]*(1-x[1]*x[1]*x[1])
+	return f1*f1 + f2*f2 + f3*f3
+}
+
+func (Beale) Df(x, grad []float64) {
+	if len(x) != 2 {
+		panic("dimension of the problem must be 2")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+
+	t1 := 1 - x[1]
+	t2 := 1 - x[1]*x[1]
+	t3 := 1 - x[1]*x[1]*x[1]
+
+	f1 := 1.5 - x[0]*t1
+	f2 := 2.25 - x[0]*t2
+	f3 := 2.625 - x[0]*t3
+
+	grad[0] = -2 * (f1*t1 + f2*t2 + f3*t3)
+	grad[1] = 2 * x[0] * (f1 + 2*f2*x[1] + 3*f3*x[1]*x[1])
+}
+
+func (Beale) Minima() []Minimum {
+	return []Minimum{
+		{
+			X:      []float64{3, 0.5},
+			F:      0,
+			Global: true,
+		},
+	}
+}
+
+// BiggsEXP2 implements the the Biggs' EXP2 function.
+//
+// Standard starting point:
+//  [1, 2]
+//
+// Reference:
+//  Biggs, M.C.: Minimization algorithms making use of non-quadratic properties
+//  of the objective function. IMA J Appl Math 8 (1971), 315-327; doi:10.1093/imamat/8.3.315
+type BiggsEXP2 struct{}
+
+func (BiggsEXP2) F(x []float64) (sum float64) {
+	if len(x) != 2 {
+		panic("dimension of the problem must be 2")
+	}
+
+	for i := 1; i <= 10; i++ {
+		z := float64(i) / 10
+		y := math.Exp(-z) - 5*math.Exp(-10*z)
+		f := math.Exp(-x[0]*z) - 5*math.Exp(-x[1]*z) - y
+		sum += f * f
+	}
+	return sum
+}
+
+func (BiggsEXP2) Df(x, grad []float64) {
+	if len(x) != 2 {
+		panic("dimension of the problem must be 2")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+
+	for i := range grad {
+		grad[i] = 0
+	}
+	for i := 1; i <= 10; i++ {
+		z := float64(i) / 10
+		y := math.Exp(-z) - 5*math.Exp(-10*z)
+		f := math.Exp(-x[0]*z) - 5*math.Exp(-x[1]*z) - y
+
+		dfdx0 := -z * math.Exp(-x[0]*z)
+		dfdx1 := 5 * z * math.Exp(-x[1]*z)
+
+		grad[0] += 2 * f * dfdx0
+		grad[1] += 2 * f * dfdx1
+	}
+}
+
+func (BiggsEXP2) Minima() []Minimum {
+	return []Minimum{
+		{
+			X:      []float64{1, 10},
+			F:      0,
+			Global: true,
+		},
+	}
+}
+
+// BiggsEXP3 implements the the Biggs' EXP3 function.
+//
+// Standard starting point:
+//  [1, 2, 1]
+//
+// Reference:
+//  Biggs, M.C.: Minimization algorithms making use of non-quadratic properties
+//  of the objective function. IMA J Appl Math 8 (1971), 315-327; doi:10.1093/imamat/8.3.315
+type BiggsEXP3 struct{}
+
+func (BiggsEXP3) F(x []float64) (sum float64) {
+	if len(x) != 3 {
+		panic("dimension of the problem must be 3")
+	}
+
+	for i := 1; i <= 10; i++ {
+		z := float64(i) / 10
+		y := math.Exp(-z) - 5*math.Exp(-10*z)
+		f := math.Exp(-x[0]*z) - x[2]*math.Exp(-x[1]*z) - y
+		sum += f * f
+	}
+	return sum
+}
+
+func (BiggsEXP3) Df(x, grad []float64) {
+	if len(x) != 3 {
+		panic("dimension of the problem must be 3")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+
+	for i := range grad {
+		grad[i] = 0
+	}
+	for i := 1; i <= 10; i++ {
+		z := float64(i) / 10
+		y := math.Exp(-z) - 5*math.Exp(-10*z)
+		f := math.Exp(-x[0]*z) - x[2]*math.Exp(-x[1]*z) - y
+
+		dfdx0 := -z * math.Exp(-x[0]*z)
+		dfdx1 := x[2] * z * math.Exp(-x[1]*z)
+		dfdx2 := -math.Exp(-x[1] * z)
+
+		grad[0] += 2 * f * dfdx0
+		grad[1] += 2 * f * dfdx1
+		grad[2] += 2 * f * dfdx2
+	}
+}
+
+func (BiggsEXP3) Minima() []Minimum {
+	return []Minimum{
+		{
+			X:      []float64{1, 10, 5},
+			F:      0,
+			Global: true,
+		},
+	}
+}
+
+// BiggsEXP4 implements the the Biggs' EXP4 function.
+//
+// Standard starting point:
+//  [1, 2, 1, 1]
+//
+// Reference:
+//  Biggs, M.C.: Minimization algorithms making use of non-quadratic properties
+//  of the objective function. IMA J Appl Math 8 (1971), 315-327; doi:10.1093/imamat/8.3.315
+type BiggsEXP4 struct{}
+
+func (BiggsEXP4) F(x []float64) (sum float64) {
+	if len(x) != 4 {
+		panic("dimension of the problem must be 4")
+	}
+
+	for i := 1; i <= 10; i++ {
+		z := float64(i) / 10
+		y := math.Exp(-z) - 5*math.Exp(-10*z)
+		f := x[2]*math.Exp(-x[0]*z) - x[3]*math.Exp(-x[1]*z) - y
+		sum += f * f
+	}
+	return sum
+}
+
+func (BiggsEXP4) Df(x, grad []float64) {
+	if len(x) != 4 {
+		panic("dimension of the problem must be 4")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+
+	for i := range grad {
+		grad[i] = 0
+	}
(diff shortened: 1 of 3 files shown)


vladimir-ch:
PTAL, @btracey

testopt/testopt.go
```diff
@@ -0,0 +1,132 @@
...
+type FunctionGradient interface {
+	FDf(x, grad []float64) float64
+}
+
+const (
+	defaultTol     = 1e-8
+	defaultGradTol = 1e-4
+)
+
+func testFunction(f Function, tests []funcTest, t *testing.T) {
```
btracey: In addition to (instead of?) wantGrad, maybe we should also test a couple of random points and compare with gonum/diff/fd.Gradient. For wantGrad, it's tempting to choose "simple" x locations like 0 and 1, which are the most likely to cause false positives. It's also a way to test the higher-dimensional functions where we have no interest in explicitly writing the gradient. fd.Gradient can be executed concurrently if we're worried about testing time.

btracey:
Is the plan to move TestMinimize (hopefully renamed to TestLocal) and the other UnconstrainedTest lists into this package as well?

testopt/testopt.go
btracey: To be clear, I'm thinking of this as a second loop after the first range.

testopt/testopt.go
```diff
@@ -0,0 +1,132 @@
...
+					t.Errorf("Test %d: %v gradient given by FDf() is incorrect. |grad - wantGrad|_∞ = %v",
+						i, reflect.TypeOf(f), dist)
+				}
+			}
+		}
+	}
+
+	if minima := f.Minima(); minima != nil {
+		for i := range minima {
+			if minima[i].X == nil {
```
btracey: I see that this case is used, but it's weird. You know what the minimum is, but you don't know where it is? Is there any utility in listing the minimum in that case?

testopt/testopt.go
```diff
@@ -0,0 +1,132 @@
...
+						i, reflect.TypeOf(f), dist)
+				}
+			}
+
+			fFunctionGradient, isFunctionGradient := f.(FunctionGradient)
+			if isFunctionGradient {
+				grad := make([]float64, len(test.wantGrad))
+				F := fFunctionGradient.FDf(test.x, grad)
+
+				if !floats.EqualWithinAbs(F, test.wantF, defaultTol) {
```
btracey: Personally, I prefer using just math.Abs() > defaultTol. Seems simpler than going to the floats function.

btracey:
What are your thoughts on exporting the testFunction to allow an easy check for outsiders with functions? It may be useful, but on the other hand maybe we don't want to guarantee anything (and maybe this isn't robust enough for complicated functions).

I think it's worth including a doc comment for this package that says something along the lines of "This is a testing yadda yadda. We encourage outside contributions of additional test functions that exhibit properties not already covered in the testing suite." (maybe something extra about 'or with significance', due to use in benchmark cases or whatever)

btracey:
Thanks a ton for putting together so many test functions. It's really important for the package.

testopt/testopt.go
vladimir-ch: Using fd.Gradient() and some random locations has crossed my mind. The locations I test now are the standard starting locations and the optimal locations.

I will try to add fd.Gradient() at random locations, but it might turn out as tricky to do in a general case. Some functions grow so fast away from a minimum that the numerical gradient can be very inaccurate. I'll see.

btracey: I've found fd.Gradient works surprisingly well,  even for function values
up in the 10^11 range. You could also specify an  [][2]float64 with  min/max

testopt/testopt.go
vladimir-ch: X can be nil in two cases:
1) the source paper lists the minimum value but not the location (possibly because it is not known analytically and listing the (inaccurate) numbers is too troublesome) and I haven't gotten around to run our optimizers to get it. Those few cases I will fix and provide also the location.
2) Some functions (from the implemented ones only two, if I am not mistaken) have minima on a line or plane. In that case I list at least the optimal value.

testopt/testopt.go
vladimir-ch: OK, I will change it.

vladimir-ch:
Moving TestLocal() to testopt: I didn't consider it, but perhaps it should be where the test cases are. If we move them and export in testopt, then TestLocal() could be there as well.

Exporting testFunction(): good idea, especially if we compare the computed gradient with fd.Gradient(), that would be very useful. I am for exporting it (after a proper revision of its signature and the tolerance checks inside it). I will add an issue.

Right now I am driven by the curiosity about why the optimization from a nearly optimum sometimes fails. I wonder if some of the failures were caused by mistakes in the test functions, when the function values were not consistent with the gradient.

I will try to add some doc comments as you suggest (too bad you wrote yadda yadda, otherwise I would just copy/paste it :-)

testopt/testopt.go
btracey: It seems like we should always enforce x != nil. It doesn't help test anything, and it's somewhat confusing. The examples that fall into case 1) we can fix, and the examples that fall into case 2) we can return any of the values that lie along the line/plane. Other points returned can check that they are sufficiently close in function value and/or that the gradient is sufficiently small.

vladimir-ch: 3) Another case is when the function is variably dimensioned and the minimum is independent of the dimension (ExtendedPowelSingular, ExtendedRosenbrock, Trigonometric)

Re 1) The only case is the Watson function at dim = 12. Our minimizers find a minimum, but the function value does not agree with the one given in More, Garbow, Hillstrom. I will add a TODO.

btracey:
If we move TestLocal() etc. here and export TestFunction(), then there is a purpose clash. Does this package test optimizers or does it test objective functions? I propose that we rename this package to "functions". Then a user could use functions.Test() on their own, or take functions.Beale{} as an example function. For testing optimizers we have a second package that is testopt, which could use the functions provided in functions (or we could choose to keep the testing suite private). This way seems to keep reasonably sized packages with one idea in mind.

With this change, I think the requirement on having Minima should be dropped, instead replaced by a Minimimumer() interface that is type asserted. Most user functions will have no idea on what/where their minima are, and implementing Minimumer is not required for using Local.

testopt/testopt.go
btracey: It's not true for ExtendedRosenbrock. There are multiple local minima in the higher dimensions. In any event, the result is the same as far as an optimizer is concerned -- the true test for finding a minima is finding a point with a sufficiently small gradient. The function value tests are only useful for confirming you have implemented the problem correctly.

With that in mind, if we have implemented the function properly and have found a point with zero gradient, then we have found an unlisted minimum.

vladimir-ch:
> If we move TestLocal() etc. here and export TestFunction(), then there is a purpose clash. Does this package test optimizers or does it test objective functions?

True.

> I propose that we rename this package to "functions". Then a user could use functions.Test() on their own, or take functions.Beale{} as an example function.

Very much agreed.

> For testing optimizers we have a second package that is testopt, which could use the functions provided in functions (or we could choose to keep the testing suite private). This way seems to keep reasonably sized packages with one idea in mind.

At the moment I would keep it private until the optimizers and tests settle down.

> With this change, I think the requirement on having Minima should be dropped, instead replaced by a Minimimumer() interface that is type asserted. Most user functions will have no idea on what/where their minima are, and implementing Minimumer is not required for using Local.

OK

testopt/testopt.go
vladimir-ch: > It's not true for ExtendedRosenbrock. There are multiple local minima in the higher dimensions.

True. I was thinking about a global minimum, the three functions I mentioned have a global minimum with a value of 0, for ExtendedRosenbrock its location is (1, ..., 1).

> ... then we have found an unlisted minimum.

Or a false minimum where the gradient is very small and finite precision does not allow further progress. It is possible that this is what is happening with the Watson function.

I have no problem with enforcing x != nil, I will modify the code accordingly.

vladimir-ch:
* Added finite difference gradient evaluation which revealed one more, hopefully last, error in a function (HelicalValley).
* Added Minimumer interface
* FuncTest.X must not be nil
* Did _not_ add exported Test() function, need to think about its arguments, etc first.
* Did _not_ add evaluation at random points, need to think how to choose those points first.

PTAL, @btracey

functions/functions.go
```diff
@@ -0,0 +1,187 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package functions provides objective functions for testing optimization
+// algorithms.
+//
+// We encourage outside contributions of additional test functions that exhibit
+// properties not already covered in the testing suite or with significance due
```
btracey: or that have significance due to prior use as benchmark cases.

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+import (
+	"math"
+	"testing"
+
+	"github.com/gonum/diff/fd"
+	"github.com/gonum/floats"
+)
+
+// Function represents an objective function.
+type Function interface {
```
btracey: These interfaces can all be private until we decide on the proper API to export.

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+	Minima() []Minimum
+}
+
+// Minimum represents information about an optimal location of a function.
+type Minimum struct {
+	// X is the location of the minimum. X may not be nil.
+	X []float64
+	// F is the value of the objective function at X.
+	F float64
+	// Global indicates whether the minimum is local or global.
```
btracey: Global indicates if the location is a global minimum.

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+	Global bool
+}
+
+type FuncTest struct {
```
btracey: I think this can be private too.

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+	WantGrad []float64 // If nil, gradient is not evaluated.
+}
+
+// TODO: Decide and implement an exported testing function:
+// func Test(f Function, ??? ) ??? {
+// }
+
+const (
+	defaultTol     = 1e-8
+	defaultGradTol = 1e-5
```
btracey: There should be two gradient tolerances. One for the specifed gradient value, which is tight (1e-14 ish), and one for finite difference which is loose (1e-6).

I think the function tolerance should be increased to 1e-14. We're comparing to a supposedly known value, so we just have to prevent against floating point noise.

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+func testFunction(f Function, tests []FuncTest, t *testing.T) {
+	// Check that the function can evaluate itself (and its gradient) correctly.
```
btracey: this should use godoc style and be before the declaration.

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+	// F is the value of the objective function at X.
+	F float64
+	// Global indicates whether the minimum is local or global.
+	Global bool
+}
+
+type FuncTest struct {
+	X []float64
+
+	WantF    float64
```
btracey: Do these need the "Want" prefixes? It's clear from context, isn't it?

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+				i, test.WantF, F)
+		}
+
+		if test.WantGrad == nil {
+			continue
+		}
+
+		// Evaluate the finite difference gradient.
+		fdGrad := make([]float64, len(test.WantGrad))
+		fd.Gradient(fdGrad, func(x []float64) float64 {
```
btracey: You can just pass f.F directly

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+			t.Errorf("Test #%d: function value given by F() is incorrect. Want: %v, Got: %v",
+				i, test.WantF, F)
+		}
+
+		if test.WantGrad == nil {
+			continue
+		}
+
+		// Evaluate the finite difference gradient.
+		fdGrad := make([]float64, len(test.WantGrad))
```
btracey: You can do 

fdGrad := fd.Gradient(nil, f.F, test.X, nil)

functions/functions.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+		}
+	}
+
+	fMinima, isMinimumer := f.(Minimumer)
+	if !isMinimumer {
+		return
+	}
+
+	// Check that the listed minima are correct.
```
btracey: Can we just append the minima to the fd test? Seems like it would reduce all this code duplication.

functions/functions.go
vladimir-ch: Thanks. Done.

functions/functions.go
vladimir-ch: Done.

functions/functions.go
vladimir-ch: Done.

functions/functions.go
vladimir-ch: Done.

functions/functions.go
vladimir-ch: I added a second gradient tolerance, that was a good idea. However, the suggested tolerances are too tight. For function values I use 1e-12 (smaller values show floating point noise in ExtendedRosenbrock). For Df() gradients I use 1e-9. Some minima are simply not known accurately enough. I disabled the test for BrownAndDennis because otherwise the value would have to be larger than 1e-6.  And for the numerical gradient the tolerance is 1e-5,  BrownBadlyScaled is the culprit (with BrownAndDennis the tolerance would have to be 1e-4).

functions/functions.go
vladimir-ch: Done.

functions/functions.go
vladimir-ch: Removed Want prefix.

functions/functions.go
vladimir-ch: Great. I should have paid more attention to the documentation.

functions/functions.go
vladimir-ch: Good idea. Done.

vladimir-ch:
I have added some local minima to ExtendedRosenbrock and I have addressed all the comments. PTAL @btracey

functions/functions.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"github.com/gonum/diff/fd"
+	"github.com/gonum/floats"
+)
+
+// Function represents an objective function.
```
btracey: lowercase comment

functions/functions.go
```diff
@@ -0,0 +1,157 @@
...
+type funcTest struct {
+	X []float64
+
+	// F is the expected function value at X.
+	F float64
+	// Gradient is the expected gradient at X. If nil, it is not evaluated.
+	Gradient []float64
+}
+
+// TODO: Decide and implement an exported testing function:
```
btracey: Please have username with TODO

functions/functions.go
```diff
@@ -0,0 +1,157 @@
...
+	defaultTol       = 1e-12
+	defaultGradTol   = 1e-9
+	defaultFDGradTol = 1e-5
+)
+
+// testFunction checks that the function can evaluate itself (and its gradient)
+// correctly.
+func testFunction(f function, ftests []funcTest, t *testing.T) {
+	// Make a copy of tests because we may append to the slice.
+	// TODO: is this necessary?
```
btracey: Just take off this TODO. A better solution isn't obvious to me, and I'm not worried about the memory allocation.

btracey:
I don't like your file names. I would have assumed that the functions themselves live in "functions.go", and I have no idea what mgh refers to.

I would suggest renaming mgh.go to functions.go and rename functions.go to validate.go

LGTM with the documentation and file name fixes.

functions/functions.go
vladimir-ch: Done.

functions/functions.go
vladimir-ch: Done.

functions/functions.go
vladimir-ch: Done.

vladimir-ch:
Fixed the documentation and the file names. Thanks for your comments, @btracey .

PR #66, opened 2015-02-04
Updated BFGS to use Symmetric

btracey:


btracey:
It's a bit slower at the moment, but it'll speed back up when gonum/matrix#100 is merged.

bfgs.go
```diff
@@ -27,10 +27,13 @@ type BFGS struct {
...
-	y []float64
-	s []float64
+	y       []float64
+	yVec    *mat64.Vector
+	s       []float64
+	tmpData []float64
+	tmpVec  *mat64.Vector
 
-	invHess *mat64.Dense // TODO: Make symmetric when mat64 has symmetric matrices
+	invHess *mat64.SymDense // TODO: Make symmetric when mat64 has symmetric matrices
```
kortschak: Remove TODO.

btracey: Done. Sorry.

kortschak:
LGTM

btracey:
@jonlawlor this looks like the same error we were seeing earlier with Travis

btracey:
Sorry, this is a real failure I guess. I think the partial pass was because I merged to matrix and here very close. Looking into it.

btracey:
Nothing was wrong with this commit. The symmetric implementation in Inner is incorrect. Fixing now.

PR #68, opened 2015-02-06
Test initial location for NaN or Inf gradient values.

btracey:
Bad initial gradient values cause the line searches to fail as the Wolfe conditions are never met. Test the initial gradient to make sure it is not NaN or Inf. While here, fix the GulfRD test which has a gradient that evaluates to NaN at the optimum location. This allows the test at the optimum to be re-enabled

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -13,12 +13,18 @@ var (
 	// ErrInf signifies the initial function value is Inf.
 	ErrInf = errors.New("optimize: initial function value is Inf")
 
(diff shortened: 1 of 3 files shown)


btracey:
PTAL @vladimir-ch

vladimir-ch:
Ah, yes, GulfRD is one of the functions that have a tiny mistake (as revealed by the tests in the new functions subpackage). I guess that is what caused the NaN. Soon #63 will be merged and these tests rewritten to use it.

LGTM

PR #69, opened 2015-02-09
Update unconstrained tests to use the functions subpackage

vladimir-ch:
Also, reduce the set of tests for CG in order to reduce the time needed
for 'go test' to finish.

### DIFF
--- a/unconstrained_test.go
+++ b/unconstrained_test.go
@@ -11,1267 +11,505 @@ import (
 	"testing"
 
 	"github.com/gonum/floats"
+	funcs "github.com/gonum/optimize/functions"
 )
 
-var negInf = math.Inf(-1)
-
-// The Fletcher-Powell helical valley function
-// Dim = 3
-// X0 = [-1, 0, 0]
-// OptX = [1, 0, 0]
-// OptF = 0
-type HelicalValley struct{}
-
-func (HelicalValley) F(x []float64) float64 {
-	θ := 0.5 * math.Atan2(x[1], x[0]) / math.Pi
-	r := math.Hypot(x[0], x[1])
-
-	f1 := 10 * (x[2] - 10*θ)
-	f2 := 10 * (r - 1)
-	f3 := x[2]
-
-	return f1*f1 + f2*f2 + f3*f3
-}
-
-func (HelicalValley) Df(x, grad []float64) {
-	θ := 0.5 * math.Atan2(x[1], x[0]) / math.Pi
-	r := math.Hypot(x[0], x[1])
-	s := x[2] - 10*θ
-	t := 5 * s / r / r / math.Pi
-
-	grad[0] = 200 * (x[0] - x[0]/r + x[1]*t)
-	grad[1] = 200 * (x[1] - x[1]/r - x[0]*t)
-	grad[2] = 2 * (x[2] + 100*s)
-}
-
-// The Biggs' EXP2 function
-// M.C. Biggs, Minimization algorithms making use of non-quadratic properties
-// of the objective function. J. Inst. Maths Applics 8 (1971), 315-327.
-// Dim = 2
-// X0 = [1, 2]
-// OptX = [1, 10]
-// OptF = 0
-type BiggsEXP2 struct{}
-
-func (BiggsEXP2) F(x []float64) (sum float64) {
-	for i := 1; i <= 10; i++ {
-		z := float64(i) / 10
-		y := math.Exp(-z) - 5*math.Exp(-10*z)
-		f := math.Exp(-x[0]*z) - 5*math.Exp(-x[1]*z) - y
-		sum += f * f
-	}
-	return sum
-}
-
-func (BiggsEXP2) Df(x, grad []float64) {
-	for i := range grad {
-		grad[i] = 0
-	}
-	for i := 1; i <= 10; i++ {
-		z := float64(i) / 10
-		y := math.Exp(-z) - 5*math.Exp(-10*z)
-		f := math.Exp(-x[0]*z) - 5*math.Exp(-x[1]*z) - y
-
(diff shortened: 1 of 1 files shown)


btracey:
@jonlawlor is this the same failure as in the other package wrt 1.4.1?

unconstrained_test.go
```diff
@@ -11,1267 +11,397 @@ import (
...
-		f:       Penalty2{},
-		x:       []float64{0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5},
-		optVal:  2.93660e-4,
-		gradTol: 1e-6,
+		f: funcs.VariablyDimensioned{},
+		x: []float64{1.00001, 1.00001, 1.00001, 1.00001, 1.00001, 1.00001, 1.00001, 1.00001, 1.00001, 1.00001},
 	},
+}
+
+var cgTests = []unconstrainedTest{
```
btracey: Why did ExtendedPowell, Beale, etc. get removed from cgtests?

unconstrained_test.go
```diff
@@ -1503,16 +623,5 @@ func testMinimize(t *testing.T, tests []UnconstrainedTest, method Method) {
 				}
 			}
 		}
-
```
btracey: Why delete this? This is a good test.

btracey:
It seems there are a bunch of changes in the functions, not just the CG ones.

btracey:
It's probably worth keeping the CG tests and using a Short testing function. Thoughts @kortschak ?

unconstrained_test.go
vladimir-ch: ExtendedPowell has been renamed to ExtendedPowellSingular, which is included in cgTests.
Beale is included in gradientDescentTests (before it wasn't) which are tested always, cgTests are appended to them.

There were indeed other functions that I forgot to include, now they should be back.

On a side note, CG is quite troublesome to test, there are many combinations of CGVariant + StepSizer and for many tests there is always one combination that always fails. Furthermore, if the tests pass locally does not mean that they pass on Travis and the cycle modify-test_locally-commit-push-wait_for_travis-check_if_travis_failed becomes annoying rather quickly.

unconstrained_test.go
vladimir-ch: It is not deleted completely, almost every test is included twice, once from the standard location and once from a slightly perturbed minimum. Moreover, this test turns out to be the most difficult to pass, so just blindly run optimization from locations returned from minimumer.Minima() is not a kind idea, imo.

jonlawlor:
@btracey the failure looks like a hiccup on the coveralls side of things.  It failed with "Couldn't find a repository matching this job" which means it couldn't find the coveralls repo, even though 2 other jobs had done so at basically the same time.  I hit rerun and it passed.  Maybe I can make the build status not depend on the coverage.

vladimir-ch:
I returned the CG tests and on my machine the tests pass in 4s which is acceptable. Previous problems with the tests might have been caused by the mistakes in some objective functions.

unconstrained_test.go
vladimir-ch: An example is the commented out test for GulfResearchAndDevelopment, both BFGS and LBFGS fail on the first iteration.

kortschak:
If it's just <tens of seconds, leave the test in.

vladimir-ch:
Watson in dimension 10 is the function due to which CG tests take long time (~1min) to finish (one test between 3 and 6s, there are 9 tests). Without it one test finishes in 0.3s, all tests in 7s. Watson is one of the hard functions and CG can hardly handle it. I added `long` field to `unconstrainedTest` and if true, I skip it when `testing.Short()`. Is this ok?

kortschak:
SGTM

vladimir-ch:
* Marked one additional test for ExtendedRosenbrock as long because of GradientDescent optimizer.
* Added two test functions for the Dai-Yuan variant of CG

`go test` takes ~100s on my machine, `go test -short` ~4.8s.

PTAL, @btracey

unconstrained_test.go
btracey: Why is this difficult to pass? Shouldn't it just exit out at the first iteration because the location is already optimal?

vladimir-ch: The starting location is a slightly perturbed minimum that does not pass the initial convergence check. Linesearch returns "failed to converge" error on the first iteration.

vladimir-ch: It seems that there was a mistake in the GulfResearchAndDevelopment function. It is fixed in #70, so it should be reviewed and merged first. Sorry for the confusion.

vladimir-ch:
It seems that the CG tests took long because there was a mistake in the Watson function. The fix is in #71 which should be reviewed and merged first. Again, sorry for the confusion and noise.

btracey:
Want to re-enable that test and we'll get this submitted. Then the next changes can be incremental improvements rather than lots of changes in this PR.

vladimir-ch:
If by "that test" you mean the minimization from an optimal location, then I think it does not belong to `testLocal()` anymore. Before, optLoc was part of `unconstrainedTest`, which has now been removed and the function itself can report its known minima. I could put the test back, but the minima would be tested several times because for almost every function there are several tests. Harmless, but unnecessary and not elegant.

I would rather prefer to add a function `testLocalFromMinima()` that would test the minima of each function only once. It could also test minimization from perturbed minima, after which I could remove many of the existing tests where I made the perturbations ad hoc. However, such changes would increase the number of changes already in this PR, so I suggest to postpone it. Do you agree, @btracey ?

btracey:
Okay. I think these tests are good for now. It looks to me like they are testing what we need to test. 

LGTM.

vladimir-ch:
Squashed.

PR #71, opened 2015-02-10
Fix Watson function.

vladimir-ch:
It appears that there was some mistake in Watson too, the CG tests waiting in the other PR now run considerably faster. Hopefully this will be the last mistake in functions to fix. PTAL, @btracey 

Used MINPACK-2 as a reference implementation.


btracey:
I'm trusting you that these are right.

LGTM.

vladimir-ch:
> I'm trusting you that these are right.

I think they are ... hopefully :-)

Thanks.

PR #72, opened 2015-02-10
Make sure CG always returns a descent direction

vladimir-ch:
Failure to generate a descent direction near a minimum is common for CG, however, the current implementation sometimes fails to generate a descent direction even when far from the minimum. This PR tries to fix it by restarting CG whenever a non-descent direction is generated. With this fix the test for the Box3D function in the other PR could be enabled.


btracey:
LGTM

PR #73, opened 2015-02-10
Add MinimalSurface function

vladimir-ch:
The first truly interesting function to be added to functions subpackage. Taken from MINPACK-2 test problem collection, good for testing our medium/large-scale minimization capabilities (although the function itself does not seem to be difficult to minimize). We can also use it to decide at which problem dimension we should switch the default method from BFGS to LBFGS. And it is also a good example of how to implement non-trivial functions.

### DIFF
--- a/functions/minsurf.go
+++ b/functions/minsurf.go
@@ -1,0 +1,232 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package functions
+
+import (
+	"fmt"
+	"math"
+)
+
+// MinimalSurface implements a finite element approximation to a minimal
+// surface problem: determine the surface with minimal area and given boundary
+// values in a unit square centered at the origin.
+//
+// References:
+//  Averick, M.B., Carter, R.G., Moré, J.J., Xue, G.-L.: The Minpack-2 Test
+//  Problem Collection. Preprint MCS-P153-0692, Argonne National Laboratory (1992)
+type MinimalSurface struct {
+	bottom, top  []float64
+	left, right  []float64
+	origin, step [2]float64
+}
+
+// NewMinimalSurface creates a new discrete minimal surface problem and
+// precomputes its boundary values. The problem is discretized on a rectilinear
+// grid with nx×ny nodes which means that the problem dimension is (nx-2)(ny-2).
+func NewMinimalSurface(nx, ny int) *MinimalSurface {
+	ms := &MinimalSurface{
+		bottom: make([]float64, nx),
+		top:    make([]float64, nx),
+		left:   make([]float64, ny),
+		right:  make([]float64, ny),
+		origin: [2]float64{-0.5, -0.5},
+		step:   [2]float64{1 / float64(nx-1), 1 / float64(ny-1)},
+	}
+
+	ms.initBoundary(ms.bottom, ms.origin[0], ms.origin[1], ms.step[0], 0)
+	startY := ms.origin[1] + float64(ny-1)*ms.step[1]
+	ms.initBoundary(ms.top, ms.origin[0], startY, ms.step[0], 0)
+	ms.initBoundary(ms.left, ms.origin[0], ms.origin[1], 0, ms.step[1])
+	startX := ms.origin[0] + float64(nx-1)*ms.step[0]
+	ms.initBoundary(ms.right, startX, ms.origin[1], 0, ms.step[1])
+
+	return ms
+}
+
+// F returns the area of the surface represented by the vector x.
+func (ms *MinimalSurface) F(x []float64) (area float64) {
+	return ms.FDf(x, nil)
+}
+
+// FDf returns the area of the surface represented by the vector x and
+// evaluates its gradient.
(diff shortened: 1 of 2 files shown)


vladimir-ch:
Finished polishing the code. PTAL, @btracey

functions/minsurf.go
```diff
@@ -0,0 +1,245 @@
...
+func (ms *MinimalSurface) GridSize() (nx, ny int) {
+	return len(ms.bottom), len(ms.left)
+}
+
+// Steps returns the spatial step sizes of the underlying rectilinear grid.
+func (ms *MinimalSurface) Steps() (hx, hy float64) {
+	return ms.hx, ms.hy
+}
+
+// WriteGnuplot saves in Gnuplot format the surface represented by the vector
```
btracey: If we keep the function in this format, it should take in a reader, and do error checking on the writes. It seems better instead to return the surface as an XYZer of some kind (like in plot). This would enable use with Plot and we could add a function to Plot to produce a gnuplot file from an XYZer. I would suggest something like
Surface(x []float64) []XYZ {}

where XYZ is a struct{X, Y, Z float64) and implements the plot.XYZer interface

functions/minsurf_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"github.com/gonum/diff/fd"
+	"github.com/gonum/floats"
+)
+
+func TestMinimalSurface(t *testing.T) {
```
btracey: Why doesn't this use testFunction?

functions/minsurf.go
kortschak: Rather, can I suggest you fill a mat64.Dense and two slices, one for row coordinates and one for column coordinates. These can all be put into a struct that is an XYZer, is more space efficient and will work with my heatmap and contour code. Free form XYZer representation won't.

kortschak: Even better, make a type  that dynamically returns the X or Y values. Basically a float64 derived type that does func (v vector) At(i int) float64 { return -0.5 + float64(i)*float64(v) } where type vector float64.

I'm still fleshing out how I want this interface to look for the surface plotting, but if you do these you get XYZer. It needs to be more subtle because, at least with my code, an XYZer is not adequate since the X and Y values must be on a grid, which XYZer doesn't guarantee.

Maybe wrap it all up in one type that is an XYZer and add to more methods, Matrix() mat64.Matrix and Coordinates() (x, y Vector) (name to be determined, but which implements the At method above and a Len() int method - although this is strictly unnecessary since Dims, but nice to be complete).

btracey: Maybe we should just remove this until we know how it fits into the bigger plotting picture.

kortschak: I have a good interface.

```
type MeshGrid interface {
    Dims() (r, c int)
    At(r, c) float64
    XCoord(int) float64
    YCoord(int) float64
}
```

This can be wrapped up to make an XYZer if needed and would be implementable with the types I described above.

vladimir-ch: * MeshGrid seems to be rather a function defined on a grid, wouldn't GridFunction be a better name?
* It would be more natural for Dims and At to use x, y rather than y, x.

I will remove WriteGnuplot().

functions/minsurf_test.go
vladimir-ch: For two reasons
1. The only alternative implementation I have is the 20+ years old MINPACK code. I did not feel like fiddling with it to make it compile, run and extract a reference solution.
2. Even if I had a reference X, I did not want to dump those large vectors into the code. I admit that this is not a serious reason.

btracey: Okay, this is fine. If we add more functions where this is the case (and I imagine we will), we should break this off into a testGradient function.

functions/minsurf.go
kortschak: I'm going to push my WIP changes for plotting surfaces soon, you can see what I mean there.

I see your point, but it complicates the common case where a grid is defined by a matrix. Currently, that is a three line shim between mat64.Matrix and GridMesh.

vladimir-ch: Ah, ok, I see what you mean.

functions/minsurf_test.go
vladimir-ch: Strictly speaking, I shouldn't be testing if the gradient at ExactX() is small (especially at such coarse grid resolutions), because ExactX is the exact solution projected on the grid, not the exact solution to the discrete problem. I will reconsider it after the weekend.

functions/minsurf.go
kortschak: When I have finished up the contour plotter I'll make a sub branch with the GridFunction approach so they can be compared. I'm thinking for the GridFunction approach that it would have `Z(c, r int) float64`, `X(c int) float64` and `Y(r int) float64` along with `Dims() (c, r int)`.

vladimir-ch:
* Removed `WriteGnuplot()` method
* Added a comment to the test which checks that the gradient at `ExactX()` is small
* I would add a `testGradient()` function later when the usage pattern is clearer

PTAL, @btracey

vladimir-ch:
When `plot.GridFunc` interface is settled and merged, I can add a method that returns something that satisfies it.

btracey:
LGTM

PR #74, opened 2015-02-17
Add tests for minimization from (nearly) optimal locations

vladimir-ch:
This PR adds two test functions:

`testLocalFromMinima()` test if `Local()` can find the minimum when the starting location is the minimum itself. At the moment the test fails when the number of major iterations taken is greater than 0, because all minima are known with sufficient accuracy, except for BrownAndDennis function, and the iterating process should not be started at all.

`testLocalFromPerturbedMinima()` test if `Local()` can find the minimum when the starting location lies in the neighborhood of the minimum, but far enough so that the iterating process is started. This test is very difficult to pass, especially when the neighborhood is small, therefore it is used only for BFGS and LBFGS with the size of the neighborhood 1e-4. A point to consider is how to generate the perturbed starting location. A random location(s)? A fixed displacement? Something else? Currently, manually perturbed minima are included in the sets of unconstrained tests, it would be nice to replace them by calls to this function.

### DIFF
--- a/unconstrained_test.go
+++ b/unconstrained_test.go
@@ -7,6 +7,7 @@ package optimize
 import (
 	"fmt"
 	"math"
+	"math/rand"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


btracey:
Have we actually caught bugs in the optimization methods with this? We found errors in the functions themselves, and we found bugs in the optimizers because the returned location didn't match the minimum location. This latter case is now checked in optimize.Local thanks to your recent addition. There is the CG issue with being close to the gradient, but I think that's also seen just by running a lot of optimizations, which we do in the tests already. I'm not sure this has a lot of value added. Additionally, as you point out, a we'll probably need to change and/or remove it later, and there are some hacky things in there (like the reflect.TypeOf stuff). It seems like maybe this is more trouble than it's worth. Sorry if I was misleading in the other PR about wanting this, you convinced me it wasn't doing much.

vladimir-ch:
Sure, no problem, I didn't like the tests much either.

Closing.

PR #77, opened 2015-02-17
Fix infinite loop when Linesearch cannot make progress.

### LINKED ISSUE #76: Linesearch with GradientDescent+Backtracking does not terminate under certain circumstances
Due to floating-point arithmetic, when getting close to a minimum of some functions, Backtracking eventually generates a step for which `xNext == ls.initX` in `Linesearch`. In such situation, Armijo condition holds, the step is accepted, MajorIteration is signalled, next descent direction (identical to the previous one) is generated and the process repeats leading to an infinite loop.

vladimir-ch:
Fixes #76

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -32,6 +32,11 @@ var (
 
 	// ErrZeroDimensional signifies an optimization was called with an input of length 0.
 	ErrZeroDimensional = errors.New("optimize: zero dimensional input")
+
+	// ErrNoProgress signifies that Linesearch cannot make further progress
+	// because there is no change in location after LinesearchMethod step due
+	// to floating-point arithmetic.
+	ErrNoProgress = errors.New("linesearch: no change in location after linesearch step")
 )
 
 // ErrMismatch signifies that the optimization function did not implement the
--- a/linesearch.go
+++ b/linesearch.go
@@ -97,6 +97,11 @@ func (ls *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, I
 		return NoEvaluation, NoIteration, err
 	}
 	floats.AddScaledTo(xNext, ls.initX, stepSize, ls.dir)
+	// Compare the starting point for the current iteration with the next
+	// evaluation point to make sure that rounding errors do not prevent progress.
+	if floats.Equal(ls.initX, xNext) {
+		return NoEvaluation, NoIteration, ErrNoProgress
+	}
 	ls.lastEvalType = evalType
 	ls.iterType = MinorIteration
 	return evalType, ls.iterType, nil
@@ -119,6 +124,11 @@ func (ls *Linesearch) initNextLinesearch(loc *Location, xNext []float64) (Evalua
 	}
 	evalType := ls.Method.Init(lsLoc, stepsize, ls.funcInfo)
 	floats.AddScaledTo(xNext, ls.initX, stepsize, ls.dir)
+	// Compare the starting point for the current iteration with the next
+	// evaluation point to make sure that rounding errors do not prevent progress.
+	if floats.Equal(ls.initX, xNext) {
+		return NoEvaluation, NoIteration, ErrNoProgress
+	}
 	ls.lastEvalType = evalType
 	ls.iterType = MinorIteration
 	return evalType, ls.iterType, nil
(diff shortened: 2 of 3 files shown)


vladimir-ch:
This needs further thought, some tests do not pass now.

vladimir-ch:
I was checking xNext == ls.initX every time xNext was generated, but it should be checked only at each MajorIteration. The testing program:
```go
package main

import (
	"fmt"

	"github.com/gonum/optimize"
	"github.com/gonum/optimize/functions"
)

func main() {
	f := functions.BrownAndDennis{}
	s := optimize.DefaultSettings()
	p := optimize.NewPrinter()
	p.ValueInterval = 0
	s.Recorder = p
	r, err := optimize.Local(f, f.Minima()[0].X, s, &optimize.GradientDescent{})
	if err != nil {
		fmt.Println(err)
	} else {
		fmt.Println(r.X)
	}
}
```
now correctly terminates with no progress error and all tests pass.

PTAL, @btracey

btracey:
Could you add that code as its own test? The Go team has tests like TestIssue79, which seems like good policy to me.

errors.go
```diff
@@ -32,6 +32,11 @@ var (
 
 	// ErrZeroDimensional signifies an optimization was called with an input of length 0.
 	ErrZeroDimensional = errors.New("optimize: zero dimensional input")
+
+	// ErrNoProgress signifies that Linesearch cannot make further progress
+	// because the next evaluation point is equal to the current one due to
+	// floating-point arithmetic.
+	ErrNoProgress = errors.New("linesearch: no progress")
```
btracey: Could we make this more informative? "linesearch: no change in location after linesearch step"

linesearch.go
```diff
@@ -62,6 +62,11 @@ func (ls *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, I
 		return NoEvaluation, ls.iterType, nil
 	}
 	if ls.iterType == MajorIteration {
+		// Compare the starting points for the previous and next iteration to
```
btracey: This doesn't seem like the right to do this test. I think the correct location is after the floats.AddScaledTo below.

btracey: Okay, I apparently can't inline comment where code didn't change. 

I think the problem is when x == x + stepsize * dir due to floating point errors. The error is that the linesearch can't make progress even though it thinks it can. This is then a "linesearch failed to converge" error. I don't think it's necessarily an error if the location at two major iterations is the same, for example the NextDirectioner wants to linesearch in a different direction starting from the same location for whatever reason.

vladimir-ch: In fact, initially I had the test after the floats.AddScaledTo because that seemed like the obvious right place. But it caused some failures in the test suite, so I thought that it was not so obvious, because without this new check the tests completed fine. I will look into it. I suspect that the explanation is the following: Backtracking produces monotonically decreasing sequence of step sizes, so once x==x+step*dir, it will not recover. However, Bisection can increase the step size, so x==x+step*dir does not mean that it cannot recover on the next minor iteration. Apparently, it can. So after floats.AddScaledTo does not seem like the right place either.

vladimir-ch:
I will add the test that you suggest, it is a good idea. I had considered it too, but then somehow neglected it.

vladimir-ch:
So it is not my subjective impression after all, @jonlawlor . In this PR, travis-ci/push failed for 1.4.1 with
```
Bad response status from coveralls: 422 - {"message":"Couldn't find a repository matching this job.","error":true}
```
while travis-ci/pr passed.

vladimir-ch:
I came to the conclusion that putting the test after floats.AddScaledTo is the right place after all. That Linesearch could overcome the "no progress due to rounding errors" situation is a coincidence thanks to how CG and (L)BFGS work. By accepting an unchanged location, the optimizer could incorporate the gradient there again into its history, so in the next iteration the descent direction was different and linesearch succeeded in finding a suitable step size. In GradientDescent the descent direction was again the same, therefore infinite loop. We loose this (not very reliable) self-correcting ability for some methods, but it can't be helped, this is the correct solution I think.

Another place where this could be dealt with is LinesearchMethod, but unlike Bisection (that keeps an interval of possible steps and can thus know if the interval is too small) a simple linesearch like Backtracking can never know if the step size gives progress or not, so delegating this check to linesearches is not acceptable.

I had to loosen the tolerance for some test problems (mostly for VariablyDimensioned function) but the tests pass and GradientDescent does not end up in infinite loop (as tested in `TestIssue76()`).

PTAL, @btracey

btracey:
LGTM

PR #78, opened 2015-02-17
Update and Go-ify docs for Settings.

### LINKED ISSUE #42: Update the docs for Settings.GradientAbsTol

vladimir-ch:
Also, fix and reformat comment for resize().

Closes #42

### DIFF
--- a/types.go
+++ b/types.go
@@ -124,36 +124,51 @@ type functionInfo struct {
 //
 // If Recorder is nil, no information will be recorded.
 type Settings struct {
-	UseInitialData       bool      // Use supplied information about the conitions at the initial x.
+	UseInitialData       bool      // Use supplied information about the conditions at the initial x.
 	InitialFunctionValue float64   // F(x) at the initial x.
 	InitialGradient      []float64 // Df(x) at the initial x.
 
-	// Converge if the objective function is less than this value.
+	// FunctionAbsTol is the threshold for acceptably small values of the
+	// objective function. FunctionAbsoluteConvergence status is returned if
+	// the objective function is less than this value.
+	// The default value is -inf.
 	FunctionAbsTol float64
 
-	// Loosely, converge if the 'average' value of the gradient is less than this
-	// value. Specifically, converge if ||grad||_2 / sqrt(len(grad)) is less than
-	// this value.
+	// GradientAbsTol determines the accuracy to which the minimum is found.
+	// GradientAbsoluteConvergence status is returned if the infinity norm of
+	// the gradient is less than this value.
 	// Has no effect if gradient information is not used.
+	// The default value is 1e-6.
 	GradientAbsTol float64
 
-	// Converge if the number of major iterations equals or exceeds this value.
+	// MajorIterations is the maximum number of iterations allowed.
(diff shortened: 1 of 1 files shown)


vladimir-ch:
Why is this coveralls thing constantly failing?

jonlawlor:
The returned message is:
```
Bad response status from coveralls: 503 - <html><body><h1>503 Service Unavailable</h1>
No server is available to handle this request.
</body></html>
```
And if I visit their website at https://coveralls.io/ I get the same 503.  They're having an outage.

I've considered having travis not fail if the coveralls code fails, but it would make it very easy for us to not notice that things weren't working on coveralls for a long period (due to issues such as this one).

This is the first outage that I've seen in ~9 months of watching.  If it is unacceptably long, or if they keep having them, we should probably eliminate it altogether.  @btracey @kortschak any thoughts as to what that point would be?

jonlawlor:
Coveralls.io is back up and I reran the builds. Things look ok.

jonlawlor:
Coveralls.io is back up and I reran the builds. Things look ok.

vladimir-ch:
Great, thanks for looking into it, @jonlawlor . I often (my subjective impression) see a failure with coveralls with the 1.4.1 build, but it seems that this was not related to it.

jonlawlor:
I get notified when there are build errors but I don't usually check them all - I assume that it is a part of normal development.  The next time you see inconsistent outcomes, where some versions of go pass and others fail, I'd appreciate it if you'd ping me with github's @ name thing.  I'll be sure to take a look.

jonlawlor:
Ah, I've just found an example while getting ready to change the 1.4.1 builds to 1.4.2.

types.go
```diff
@@ -121,36 +121,51 @@ type functions struct {
...
+	// FunctionAbsTol is the threshold for acceptably small values of the
+	// objective function. FunctionAbsoluteConvergence status is returned if
+	// the objective function is less than this value.
+	// The default value is -inf.
 	FunctionAbsTol float64
 
-	// Loosely, converge if the 'average' value of the gradient is less than this
-	// value. Specifically, converge if ||grad||_2 / sqrt(len(grad)) is less than
-	// this value.
+	// GradientAbsTol determines the accuracy with which the minimum is to be
```
btracey: to which the minimum is found.

types.go
```diff
@@ -121,36 +121,51 @@ type functions struct {
...
 	// If it equals zero, this setting has no effect.
+	// The default value is 0.
 	MajorIterations int
 
-	// Converge if the duration of the run is longer than this value. Runtime
-	// is only checked at iterations of the optimizer. If it equals zero,
-	// this setting has no effect.
+	// Runtime is the maximum runtime allowed. RuntimeLimit status is returned
+	// if the duration of the run is longer than this value. Runtime is only
+	// checked at iterations of the optimizer.
```
btracey: this should be "iterations of the Method", right?

btracey:
LGTM after the minor fixes.

vladimir-ch:
Fixed both comments.

kortschak:
@jonlawlor can you make it retry on failure (once only) for coveralls?

PR #80, opened 2015-02-23
Rename functions struct and embed FunctionInfo in it

vladimir-ch:
Benefits:

* Type switched methods are together with bools that indicate capabilities of the objective function
* Number of arguments to various internal functions can be reduced by one
* Removes the name conflict with the functions subpackage


btracey:
LGTM

PR #81, opened 2015-02-26
Don't return NotTerminated from minimize()

vladimir-ch:
This has been discussed before and I think that it is indeed better not to return NotTerminated status, just as you suggested @btracey .


btracey:
LGTM

PR #82, opened 2015-02-26
Rename function interface

vladimir-ch:
This PR is proposal how the existing function interfaces could be modified in order to better accomodate functions that can evaluate their Hessian. This prepares the way for an implementation of Newton's method.

At present an objective function can implement the following three methods:
```go
F() // function value
Df() // gradient
FDf() // function value and gradient
```
For the Hessian the natural progression would be
```go
DDf(hess *mat64.SymDense) // hessian
FDfDDf(x float64, grad []float64, hess *mat64.SymDense) // function value, gradient and hessian
```
which I don't like much. Moreover, there is no naming connection to the corresponding EvaluationType and values in Stats and Settings.

Alternatively, the Hessian method could be
```go
Hessian(hess *mat64.SymDense)
```
or
```go
Hess(hess *mat64.SymDense)
```
but in both cases there is the question of how the three-argument method would be called. In any case, this is incosistent with the existing methods.

Yet another alternative is this PR. The interface methods are:
```go
Func()
Grad()
Hess()
FuncGrad()
FuncGradHess()
```
the corresponding evaluation types:
```go
FuncEvaluation
GradEvaluation
HessEvaluation
FuncGradEvaluation
FuncGradHessEvaluation
```
and similarly for Stats and Settings (`FuncEvaluations`, ...), making thus a clear link between methods, evaluation requests from optimizers, settings (limits how many times a method can be called) and stats (how many times a method has actually been called).

PTAL, @btracey @kortschak

### DIFF
--- a/backtracking.go
+++ b/backtracking.go
@@ -56,7 +56,7 @@ func (b *Backtracking) Init(loc LinesearchLocation, step float64, _ *FunctionInf
 	b.stepSize = step
 	b.initF = loc.F
 	b.initG = loc.Derivative
-	return FunctionEval
+	return FuncEvaluation
 }
 
 func (b *Backtracking) Finished(loc LinesearchLocation) bool {
@@ -68,5 +68,5 @@ func (b *Backtracking) Iterate(_ LinesearchLocation) (float64, EvaluationType, e
 	if b.stepSize < minimumBacktrackingStepSize {
 		return 0, NoEvaluation, ErrLinesearchFailure
 	}
-	return b.stepSize, FunctionEval, nil
+	return b.stepSize, FuncEvaluation, nil
 }
--- a/bisection.go
+++ b/bisection.go
@@ -53,7 +53,7 @@ func (b *Bisection) Init(loc LinesearchLocation, step float64, _ *FunctionInfo) 
 	b.minGrad = loc.Derivative
 	b.maxGrad = math.NaN()
 
-	return FunctionAndGradientEval
+	return FuncGradEvaluation
 }
 
 func (b *Bisection) Finished(loc LinesearchLocation) bool {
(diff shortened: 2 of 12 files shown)


kortschak:
I think this is an improvement.

btracey:
I agree

btracey:
Let us know when this is in a state you want reviewed.

vladimir-ch:
I think this can be reviewed. PTAL, @btracey @kortschak .

kortschak:
LGTM but wait fro @btracey.

Nothing changed since 5 days ago?

vladimir-ch:
I wanted to check the code to see if there was anything else that should go into this PR, but didn't have time to do so until yesterday. There are things that will have to be done but I didn't want to increase the number of changes in this PR:

1. Add `HessEvaluations` and `FuncGradHessEvaluations` to `Stats`
2. Add `HessEvaluations` to `Settings`
3. Add `HessianEvaluationLimit` status. In fact, I think it would be nice to call the statuses that correspond to `MajorIterations`, `FuncEvaluations`, `GradEvaluations` and `HessEvaluations` by appending `Limit`
4. Add `IsHessian` and `IsFunctionGradientHessian` to `FunctionInfo` and the extracted methods to `functionInfo`

On the other hand, it would be natural to include 1.-3., so I will do it. I would leave 4. for later and update `getFunctionInfo()` at the same time.

kortschak:
OK SGTM

btracey:
We should keep this change isolated, and the additions next. I'll try to get to the reveiw soon.

vladimir-ch:
OK

btracey:
Could you remove the Hessian references (FuncGradHess, Hessian interface, etc.)? That was this PR is just the name change to Func and Grad, and then the next PR can add the Hessian information.

btracey:
LGTM after that.

vladimir-ch:
I removed the last two commits and will submit them in a new PR.

Thanks.

PR #83, opened 2015-03-04
Add Hessian interface

vladimir-ch:
Included two commits from #82 and addressed points 1-3 from the discussion therein. PTAL, @btracey .

As I mentioned, we could consider renaming the four statuses so that their name matches the corresponding fields in Settings, e.g. `IterationLimit` to `MajorIterationsLimit`, `HessianEvaluationLimit` to `HessEvaluationsLimit`, ...

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -4,6 +4,8 @@
 
 package optimize
 
+import "github.com/gonum/matrix/mat64"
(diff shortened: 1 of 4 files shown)


btracey:
LGTM

PR #84, opened 2015-03-09
Add Method.Requires()

vladimir-ch:


vladimir-ch:
This PR together with #85 and #86 conclude the Hessionization of the core framework. PTAL, @btracey 

The next step will be to correctly issue evaluation requests at SubIterations from Linesearch and then a Newton optimizer can finally be added.

bfgs.go
```diff
@@ -159,3 +159,7 @@ func (b *BFGS) NextDirection(loc *Location, dir []float64) (stepSize float64) {
 	floats.Scale(-1, dir)
 	return 1
 }
+
+func (_ BFGS) Requires() (grad, hess bool) {
```
btracey: You don't need the underscore, just func(BFGS). It should be func(*BFGS) though to be consistent. This change goes for the other optimizers.

vladimir-ch: I was not sure if it was better to include _ or not. Omitting * was my mistake though. Will fix soon.

vladimir-ch:
PTAL, @btracey

interfaces.go
```diff
@@ -79,6 +79,9 @@ type Method interface {
 
 	// Stores the next location to evaluate in xNext
 	Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error)
+
+	// Requires returns the minimum requirements of Method on Function.
```
btracey: I think this could use a more elaborate comment. For example, there isn't a direct correspondance between the requirements and the methods (you can have FuncGrad without Grad). It's tricky enough that a couple sentences are okay. "Requires specifies information about the objective function needed by the optimizer beyond just the function value. For example, some optimizers must have the gradient ... "

types.go
```diff
@@ -121,6 +121,15 @@ type functionInfo struct {
 	statuser         Statuser
 }
 
+func (f functionInfo) canEvaluate(grad, hess bool) error {
```
btracey: Could you add: "TODO(btracey): Think about making this an exported function when the constraint interface is designed"
It would be preferable to allow users to check without calling Local, but until we have the full story it may as well be private.

types.go
```diff
@@ -121,6 +121,15 @@ type functionInfo struct {
...
+	if grad && !(f.IsGradient || f.IsFunctionGradient) {
+		return fmt.Errorf("function does not implement Grad() nor FuncGrad()")
```
btracey: this can just be errors.New()

Should be "does not implement Gradient or FunctionGradient", as those are the interface names.

vladimir-ch: A grammar question: should there be "or" or "nor"?

kortschak: "either" -> "or"
"neither" -> "nor"

So in this case "or" (equivalent here to `!(a||b)`).

interfaces.go
vladimir-ch: My intention for `Requires()` was to just provide information what kind of optimizer the Method is (derivative-free, gradient-based, ...), without considering how this information is used and how it is translated into interfaces that Function should satisfy. That's why I kept the comment rather minimal, because `Requires()` is of interest only to implementers of external minimizers and to us, not to users of optimize.

On the other hand, matching the requirements of Method and capabilites of Function is tricky and should be clearly explained in the documentation because it directly affects the users of optimize, it tells them what interfaces Function should satisfy. But `Requires()` is not the right place for such explanation, I would say that `Local()` or the package docs would be a better place. I think that evaluations returned by `Init()` and `Iterate()` are more important for the matching between requirements and the methods of Function, maybe each optimizer should indicate the evaluations it requests in its docs.

I will try to improve the comment for `Requires()` anyway.

types.go
vladimir-ch: I see. I was just (wrongly) wondering if that "not" somehow implicitly substitutes "neither". Thanks for the explanation.

interfaces.go
vladimir-ch: I mean, `Requires()` is just a summary of what `Init()` and `Iterate()` return during the run. The matching of EvaluationTypes to what Function implements takes place even if `Requires()` does not exist, so we should clearly document it anyway. I think `Function` will be the best place for it since anyone calling `Local()` will look it up in order to implement their objective function.

vladimir-ch:
As always, great and helpful comments, @btracey . Could you PTAL?

types.go
kortschak: Trending off topic. You could write, "does not implement Grad(), nor FuncGrad()", but it's beginning to sound dramatic, almost Shakspearean.

interfaces.go
btracey: > On Mar 11, 2015, at 11:22 PM, Vladimír Chalupecký <notifications@github.com> wrote:
> 
> In interfaces.go <https://github.com/gonum/optimize/pull/84#discussion_r26280947>:
> 
> > @@ -79,6 +79,9 @@ type Method interface {
> >  
> >  	// Stores the next location to evaluate in xNext
> >  	Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error)
> > +
> > +	// Requires returns the minimum requirements of Method on Function.
> My intention for Requires() was to just provide information what kind of optimizer the Method is (derivative-free, gradient-based, ...), without considering how this information is used and how it is translated into interfaces that Function should satisfy. That's why I kept the comment rather minimal, because Requires() is of interest only to implementers of external minimizers and to us, not to users of optimize.
> 
The way the package is designed is that “users of optimize” is both those who want their functions minimized, and who want to use the “framework” of local when implementing a new optimizer. This is why Method is a publicly exported interface instead of it being a list of possible optimizers.  The Method documentation should thus speak to those who want to implement an algorithm.

You are right about a decoupling from the “interfaces that Function should satisfy”. Mostly, I thought the comment was a little spartan. When I get a chance I’ll try to say what I mean taking your comments into consideration.
> On the other hand, matching the requirements of Method and capabilites of Function is tricky and should be clearly explained in the documentation because it directly affects the users of optimize, it tells them what interfaces Function should satisfy. But Requires() is not the right place for such explanation, I would say that Local() or the package docs would be a better place.
> 
Agreed on both accounts
> I think that evaluations returned by Init() and Iterate() are more important for the matching between requirements and the methods of Function, maybe each optimizer should indicate the evaluations it requests in its docs.
> 

Interesting point. I’ll ponder.

kortschak:
Late in the day, but I was just thinking about this some more and a couple of things came to me. Neither of these is something that must be acted on, just to be considered. Neither would be a huge change from the current state of the PR.

1. It is possible to use optional interfaces to do this rather than add Requires. The demand from the method then is determined by whether the method satisfies the demand interface.
2. (bike-sheddy) Naming in Go idiom tends towards the Anglo-Saxon. This reduces the syllable count and makes the code more friendly (at least to native English speakers). I think this probably is a result of rsc's views on names. Two alternatives to Requires in this vein are Needs and Wants (depending on the severity of the requirement).

If the two are incorporated, you would end up with something like

```
func (*LBFGS) NeedsGrad() {}
```

This gives a nice API and documentation effect (with the possibility of adding explanations for why the extra information is needed in the optimisation iteration - I think we have an opportunity in gonum to be using the documentation for teaching and not just research), while dropping back the number of methods in `optimize.Method`.

vladimir-ch:
I must say I like your suggestion a lot, @kortschak . The generic `optimize.Method` interface stays minimal. What a particular optimizer needs is a part of its type/interface. It would nicely split the documentation: e.g. `NeedsGrader`'s documentation would be for implementers of outside optimizers, while `func NeedsGrad()` documentation in a particular optimizer would be for callers of `Local()` so that they know that their `Function` must be able to evaluate the gradient. Similarly for the Hessian and constraints.

If this optional interfaces approach is adopted, then I would be for doing so already in this PR, we could avoid writing a proper comment for `Requires()` that would go away anyway. What does @btracey think, especially wrt constraints?

btracey:
I am tepid on optional interfaces. First of all, explicit is better than implicit, and so where possible one should prefer explicit interface requirements. Secondly, optional interfaces without return make composition harder. I have been looking at some simple wrappers around a Function. One of the challenges is that, as far as I can tell, it's impossible for the outer type implement Gradient if and only if the Function it contains implements gradient. 

For the Function types, the method signaling approach seems like an unfortunate necessity. My previous attempts at the optimize design showed how bad the combinatorics can get, and it would be over-burdensome to force every function to implement a lot of methods before use. This is especially true since new functions will be frequently created by users.

I don't think optional interfaces are necessary for optimization methods. Optimization methods are more "permanent" (i.e. they are frequently used for more than one Function), and thus the burden of extra methods gets amortized. I believe there's also a constrained set of information we need from the optimizer. The three methods are:

````
// IsLocal returns true if the method is for local optimization (as opposed to global optimization)
IsLocal() bool

// Needs declares the information from the optimization Function needed by the Method.
Needs() struct {
      Gradient bool    // this would be a named type 
      Hessian bool 
      Bounds bool
}

// Allows declares the constraint sets supported by the optimization method
Allows() struct{
    Bounds bool
    LinearEquality bool
    LinearInequality bool
    NonlinearEquality bool
    NonlinearEquality bool
}
````
Three methods are quite a bit, but it allows for good error checking by Local (and the future Global). These methods also double as documentation for the optimization method.

**Comment by kortschak** [comment]
>One of the challenges is that, as far as I can tell, it's impossible for the outer type implement Gradient if and only if the Function it contains implements gradient. 

This is true, unless the client adds the type marker to their type.

I think overall, you are right - I disagree on some issues, but the overall feeling comes down on your side of the fence.

As a side comment, I think the three interfaces you propose, which do seem to cover the spectrum of concerns are probably the only time I've seen where an unnamed type return actually makes sense (yes, I know you say they should be named - I'm not so sure; the signature very nicely self-documents locally rather than needing to navigate to somewhere else in the godoc).

**Comment by btracey** [comment]
> This is true, unless the client adds the type marker to their type.
Could you explain this?

It may make sense to return an unnamed type. As you say, the self documentation is very nice. It's slightly more annoying for the Method implementer as they have to construct the type and fill it in, but it's only once per Method and copy/paste should be easy.

**Comment by kortschak** [comment]
>Could you explain this?

You are right that an embedded interface can not be asserted to an interface that is broader than the embedded interface unless the embedding type implements the methods needed to implement the interface being asserted. You can add those extra methods explicitly though.

The following will not compile:
```
type userType struct {
    Method
    // other things
}

type GradNeeder interface {
    NeedsGrad()
}

var _ GradNeeder = userType{}
```

But the client could add the NeedsGrad method - this bit is why I fall on your side of the fence; it's unnecessarily onerous if userType has exactly the same behaviour wrt needs as the embedding type (which many may).

> It's slightly more annoying for the Method implementer as they have to construct the type and fill it in, but it's only once per Method and copy/paste should be easy.

And if they want they can define a named type for it.

**Comment by btracey** [comment]
SGTM

**Comment by vladimir-ch** [comment]
`IsLocal()` and `Allows()` are not needed at the moment, so I will focus on `Needs()`.

The unnamed struct looks nice in the interface definition, but for implementers it is indeed rather annoying:

```go
func (*BFGS) Needs() struct {
	Gradient bool
	Hessian  bool
	Bounds bool
} {
	return struct {
		Gradient bool
		Hessian  bool
		Bounds bool
	}{true, false, false}
}
```

Or we can have a named, kind of useless (more so if exported) type:

```go
type methodNeeds struct {
	Gradient bool
	Hessian  bool
	Bounds bool
}

func (*BFGS) Needs() methodNeeds {
	return methodNeeds{true, false, false}
}
```

Or we can just return three bools:

```go
func (*BFGS) Needs() (gradient, hessian, bounds bool) {
	return true, false, false
}
```

Or we can use optional interfaces:

```go
func (*BFGS) NeedsGradient() {}
```

In the struct version I like very much that one can write `if method.Needs().Gradient`. Any consensus? Unnamed structs?

And I just wanted to correctly allocate and fill the initial location...

btracey:
It's important to keep the big picture in mind.

I discussed above why I feel the struct approach is better than implicit interfaces. I think the struct is better than multiple bools. It'll be easier to move around, easier to extend if we find we missed something. At three bools it could go one way or another, but at 5 bools  in Allows, it should be a struct, and the two should agree in implementation. Dan's point about documentation is a good one. There is stutter in the code, but it is easily copied and pasted, and not all that much more as would be necessary with a named type.

vladimir-ch:
Unnamed structs are a nice solution. I updated the code, PTAL @btracey @kortschak

kortschak:
LGTM but wait for @btracey

types.go
```diff
@@ -121,6 +122,18 @@ type functionInfo struct {
 	statuser         Statuser
 }
 
+// TODO(btracey): Think about making this an exported function when the
+// constraint interface is designed.
+func (f functionInfo) satisfies(method Method) error {
+	gradOk := f.IsGradient || f.IsFunctionGradient
+	if method.Needs().Gradient && !gradOk {
+		return errors.New("function does not implement Gradient or FunctionGradient")
```
btracey: @kortschak do errors get "optimize: " prefixes?

This error needs to be a bit better. The error is that the Method needs the gradient and that the Function can not supply it.

"Method needs gradient information but Function does not implement Gradient or FunctionGradient"

btracey:
LGTM after error message change.

types.go
kortschak: Add the prefix: "optimize: function does not implement needed Gradient or FunctionGradient"

vladimir-ch:
I went with @kortschak 's version of the error message. Rebased and merged. This time I used `--no-ff`, I didn't want to squash the three commits, but I also wanted to keep them together as part of one larger step.

Thanks for the review.

PR #85, opened 2015-03-10
Add Hessian to (F|f)unctionInfo

vladimir-ch:
#84 should be reviewed and merged first.

### DIFF
--- a/types.go
+++ b/types.go
@@ -105,9 +105,11 @@ type Stats struct {
 
 // FunctionInfo is data to give to the optimizer about the objective function.
 type FunctionInfo struct {
-	IsGradient         bool
-	IsFunctionGradient bool
-	IsStatuser         bool
+	IsGradient                bool
+	IsHessian                 bool
+	IsFunctionGradient        bool
+	IsFunctionGradientHessian bool
+	IsStatuser                bool
 }
 
 // functionInfo contains information about which interfaces the objective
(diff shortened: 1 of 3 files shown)


vladimir-ch:
The main change is in `functionInfo.satisfies(Method)`. The logic is to satisfy Method by any means available, even if inefficient (e.g., using FuncGradHess() to evaluate only the gradient). Modifications to `evaluate()` in the next PR will match this logic, too.

I appended "interface" to the error messages for better clarity, but in the first three messages I am not sure if it should be in singular (only one is enough) or plural (there are two or three interfaces listed).

PTAL, @btracey @kortschak

kortschak:
Singular is fine - it's contextual so you cannot be correct here if you want to include that (unless we are using the word to describe the total interface, which is perfectly reasonable).

Number in language can be fluid.

vladimir-ch:
Thanks, I'll leave it in singular, then.

types.go
```diff
@@ -116,21 +118,68 @@ type FunctionInfo struct {
...
 
 // TODO(btracey): Think about making this an exported function when the
 // constraint interface is designed.
 func (f functionInfo) satisfies(method Method) error {
-	gradOk := f.IsGradient || f.IsFunctionGradient
-	if method.Needs().Gradient && !gradOk {
-		return errors.New("optimize: function does not implement needed Gradient or FunctionGradient")
+	if method.Needs().Gradient {
+		if method.Needs().Hessian {
+			// Method needs the gradient and the Hessian, any of these three interfaces will do.
```
btracey: I don't think this is right (or the hessian check below). If the method needs the Gradient, it can be gotten from FunctionGradientHessian. It's a bit more code in Local to do the right thing, but the information is there.

vladimir-ch: You are right. Some extra allocations will be necessary, but the information is there. This will simplify the checks, I will postpone the actual allocations and evaluation to the next PR.

vladimir-ch:
Fixed `functionInfo.satisfies()`. PTAL @btracey

btracey:
LGTM

PR #86, opened 2015-03-10
Add Hessian to Location

vladimir-ch:
#84 and #85 should be reviewed and merged first.

### DIFF
--- a/local.go
+++ b/local.go
@@ -336,10 +336,7 @@ func evaluate(funcInfo *functionInfo, evalType EvaluationType, xNext []float64, 
 	switch evalType {
 	case FuncEvaluation:
 		if different {
-			// Invalidate the gradient because it will not be evaluated.
-			for i := range loc.Gradient {
-				loc.Gradient[i] = math.NaN()
-			}
+			invalidate(loc, false, false, true, true)
 		}
 		loc.F = funcInfo.function.Func(loc.X)
 		stats.FuncEvaluations++
@@ -347,30 +344,85 @@ func evaluate(funcInfo *functionInfo, evalType EvaluationType, xNext []float64, 
 	case GradEvaluation:
 		if funcInfo.IsGradient {
 			if different {
-				// Invalidate the function value because it will not be
-				// evaluated.
-				loc.F = math.NaN()
+				invalidate(loc, false, true, false, true)
 			}
 			funcInfo.gradient.Grad(loc.X, loc.Gradient)
 			stats.GradEvaluations++
 			return
 		}
 		if funcInfo.IsFunctionGradient {
+			if different {
+				invalidate(loc, false, false, false, true)
+			}
 			loc.F = funcInfo.functionGradient.FuncGrad(loc.X, loc.Gradient)
 			stats.FuncGradEvaluations++
 			return
 		}
+		if funcInfo.IsFunctionGradientHessian {
+			if loc.Hessian == nil {
+				loc.Hessian = mat64.NewSymDense(len(loc.X), nil)
+			}
+			loc.F = funcInfo.functionGradientHessian.FuncGradHess(loc.X, loc.Gradient, loc.Hessian)
+			stats.FuncGradHessEvaluations++
+			return
+		}
+	case HessEvaluation:
+		if funcInfo.IsHessian {
+			if different {
+				invalidate(loc, false, true, true, false)
+			}
+			funcInfo.hessian.Hess(loc.X, loc.Hessian)
+			stats.HessEvaluations++
+			return
+		}
+		if funcInfo.IsFunctionGradientHessian {
+			if loc.Gradient == nil {
+				loc.Gradient = make([]float64, len(loc.X))
+			}
+			loc.F = funcInfo.functionGradientHessian.FuncGradHess(loc.X, loc.Gradient, loc.Hessian)
+			stats.FuncGradHessEvaluations++
+			return
+		}
 	case FuncGradEvaluation:
 		if funcInfo.IsFunctionGradient {
+			if different {
+				invalidate(loc, false, false, false, true)
+			}
 			loc.F = funcInfo.functionGradient.FuncGrad(loc.X, loc.Gradient)
 			stats.FuncGradEvaluations++
 			return
 		}
 		if funcInfo.IsGradient {
+			if different {
+				invalidate(loc, false, false, false, true)
+			}
 			loc.F = funcInfo.function.Func(loc.X)
 			stats.FuncEvaluations++
 			funcInfo.gradient.Grad(loc.X, loc.Gradient)
 			stats.GradEvaluations++
+			return
+		}
+		if funcInfo.IsFunctionGradientHessian {
+			if loc.Hessian == nil {
+				loc.Hessian = mat64.NewSymDense(len(loc.X), nil)
+			}
+			loc.F = funcInfo.functionGradientHessian.FuncGradHess(loc.X, loc.Gradient, loc.Hessian)
+			stats.FuncGradHessEvaluations++
+			return
+		}
+	case FuncGradHessEvaluation:
+		if funcInfo.IsFunctionGradientHessian {
+			loc.F = funcInfo.functionGradientHessian.FuncGradHess(loc.X, loc.Gradient, loc.Hessian)
+			stats.FuncGradHessEvaluations++
+			return
+		}
+		if funcInfo.IsGradient && funcInfo.IsHessian {
+			loc.F = funcInfo.function.Func(loc.X)
+			stats.FuncEvaluations++
+			funcInfo.gradient.Grad(loc.X, loc.Gradient)
+			stats.GradEvaluations++
+			funcInfo.hessian.Hess(loc.X, loc.Hessian)
+			stats.HessEvaluations++
 			return
 		}
 	case NoEvaluation:
@@ -378,13 +430,7 @@ func evaluate(funcInfo *functionInfo, evalType EvaluationType, xNext []float64, 
 			// The evaluation point xNext is not equal to loc.X, so we fill loc
 			// with NaNs to indicate that it does not contain a valid
 			// evaluation result.
-			loc.F = math.NaN()
-			for i := range loc.X {
-				loc.X[i] = math.NaN()
-			}
-			for i := range loc.Gradient {
-				loc.Gradient[i] = math.NaN()
-			}
+			invalidate(loc, true, true, true, true)
 		}
 		return
 	default:
--- a/local.go
+++ b/local.go
@@ -311,18 +311,11 @@ func invalidate(loc *Location, f, grad, hess bool) {
 	if f {
 		loc.F = math.NaN()
 	}
-	if grad {
-		for i := range loc.Gradient {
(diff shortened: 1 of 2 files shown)


vladimir-ch:
This is how adding Hessian to Location and its evaluation in `evaluate()` could look like. PTAL, @btracey .

btracey:
Couple of high-level thoughts.

1) I don't think invalidate should be a method on location. I think it should be a function.
2) My original intent behind invalidate is just to act as a safety check. Code should never rely on the values in Location being invalidated, and so strictly this step is unnecessary. It seems sad to do O(n^2) work during an unnecessary safety step. How about just making the first entry NaN (of both Gradient and Hessian). That should be enough to corrupt a buggy optimizer but does not waste a lot of time.
3) It feels like something is wrong with the package design if the Proxy fields are actually necessary.  Location represents the data values at the location. There should not be any harm in giving a Method more data than it requested. If our code needs nil values in Location to do the right thing, then we are relying on more information content in Location than it should carry. 
I did a quick glance through the code, and here is what I see:
GradientConvergence: Checks if gradient is nil. I'm not sure what the right path here is. If you are running a non-gradient-based optimizer and we find the gradient is small, should it converge?
Linesearch: This one we can fix. One option is a usesGradient boolean. I'm not sure it's actually a problem. All inputs have a Function method, so if a linesearch method is gradient-free, then it just requests function values.

vladimir-ch:
> 1) I don't think invalidate should be a method on location. I think it should be a function.

No problem, I can change that.

> 2) My original intent behind invalidate is just to act as a safety check. Code should never rely on the values in Location being invalidated, and so strictly this step is unnecessary. It seems sad to do O(n^2) work during an unnecessary safety step. How about just making the first entry NaN (of both Gradient and Hessian). That should be enough to corrupt a buggy optimizer but does not waste a lot of time.

It is not only sad, it is hideous. I always took the invalidating with NaNs as a temporary solution. The problem is that Recorders get to see the location after every evaluation and they need to know which fields are valid and which are not. With NaNs they will at least not report something incorrect as seemingly correct (e.g., every minor iteration with GradientDescent and Backtracking, or with Newton and any linesearch). Setting the first entry to NaN would be acceptable internally, but at the moment I can't imagine signalling invalidity to Recorders in this way.

One minor thing we can do is avoid invalidating the whole location when NoEvalution and xNext != loc.X as we do now. I cannot see any sane reason why an optimizer would make such a request (basically, fill the location for the next iteration with NaNs). In such a situation the execution flow in minimize() makes no sense and I would lean to panic in evaluate() instead of invalidating location. So loc.X would be always valid.

We could also simplify our life considerably if we reduced the number of evaluation types (and thus function interfaces) to just three: Function, FunctionGradient and FunctionGradientHessian. In this way, loc.F would be always valid, internal logic simplified, docs would become clearer and Local()'s caller's life easier (gradient-free method: implement only Function, gradient-based method: implement the first two, newton: implement at least the first and the third, or all three for better performance). The price would be some extra evaluations of Func and Grad.

And talking about valid and invalid locations, there is a bug in the code: when iteration or runtime limit is reached during a minor iteration, the location we return can be invalid - if at a minor iteration a lower F is found, optLoc is updated but with an incomplete data (again, any line search method with Backtracking, or any Newton method that does not update the Hessian at minor iterations).

> 3) It feels like something is wrong with the package design if the Proxy fields are actually necessary. Location represents the data values at the location. There should not be any harm in giving a Method more data than it requested. If our code needs nil values in Location to do the right thing, then we are relying on more information content in Location than it should carry. 
I did a quick glance through the code, and here is what I see:
GradientConvergence: Checks if gradient is nil. I'm not sure what the right path here is. If you are running a non-gradient-based optimizer and we find the gradient is small, should it converge?
Linesearch: This one we can fix. One option is a usesGradient boolean. I'm not sure it's actually a problem. All inputs have a Function method, so if a linesearch method is gradient-free, then it just requests function values.

I don't like the proxies either, I just wanted a code as a base for discussion. But I see what you mean and perhaps I overcomplicated the matter. If we get a gradient-based method and a function that implements only Func() and FuncGradHess(), then we have no choice but to allocate the Hessian in Location. The method will not need it or look at it, but it will be there. If we get a gradient-free method, then Gradient will never be allocated and GradientConvergence check will be false. Then there is the weird situation when method returns that it needs only the Hessian. We either say that in such a case it is assumed that gradient is needed too, or we panic, or we handle it in the code. As for Linesearch, we will have to do something with it anyway because now it does not handle Hessian evaluations (but it can wait for another PR).

Summary: I will remove the proxies from functionInfo and make invalidate() a function. I would like to panic in evaluate() when `NoEvaluation && different`, but I will wait for your comment. As for invalidating of Location, I see only NaNs or adding bool fields to Location, open for discussion. The bug with returning invalid location will go to another PR. And now I realize I must also update copyLocation().

btracey:
Agreed with panic on NoEvaluation && different. Still processing the rest.

btracey:
>>    2) My original intent behind invalidate is just to act as a safety check. Code should never rely on the values in Location being invalidated, and so strictly this step is unnecessary. It seems sad to do O(n^2) work during an unnecessary safety step. How about just making the first entry NaN (of both Gradient and Hessian). That should be enough to corrupt a buggy optimizer but does not waste a lot of time.

> It is not only sad, it is hideous. 

Truth

> I always took the invalidating with NaNs as a temporary solution. The problem is that Recorders get to see the location after every evaluation and they need to know which fields are valid and which are not. With NaNs they will at least not report something incorrect as seemingly correct (e.g., every minor iteration with GradientDescent and Backtracking, or with Newton and any linesearch). Setting the first entry to NaN would be acceptable internally, but at the moment I can't imagine signalling invalidity to Recorders in this way.

The recorders don't need such signaling. They know EvaluationType and Location, so they can know which fields are valid. Similarly, the optimizers should know what's valid, since they are the ones requesting the data. The thought of having the NaN is just to have it to help prevent silent bugs from Methods using the Gradient when they previously requested a function. If you, as an implementer, see NaNs, you know you have a bug somewhere (assuming the function doesn't NaN). 

> One minor thing we can do is avoid invalidating the whole location when NoEvalution and xNext != loc.X as we do now. I cannot see any sane reason why an optimizer would make such a request (basically, fill the location for the next iteration with NaNs). In such a situation the execution flow in minimize() makes no sense and I would lean to panic in evaluate() instead of invalidating location. So loc.X would be always valid.

SGTM

> We could also simplify our life considerably if we reduced the number of evaluation types (and thus function interfaces) to just three: Function, FunctionGradient and FunctionGradientHessian. In this way, loc.F would be always valid, internal logic simplified, docs would become clearer and Local()'s caller's life easier (gradient-free method: implement only Function, gradient-based method: implement the first two, newton: implement at least the first and the third, or all three for better performance). The price would be some extra evaluations of Func and Grad.

Yes, this sounds reasonable. In the Convex world, I'm not sure this is true, but in the non-convex world (like this package), you probably always need at least the function value. I wish there were a way to not recompute the function value, but I would guess for most functions they need to compute the function value before they can compute the gradient (or at least have a significant function overlap). 

> And talking about valid and invalid locations, there is a bug in the code: when iteration or runtime limit is reached during a minor iteration, the location we return can be invalid - if at a minor iteration a lower F is found, optLoc is updated but with an incomplete data (again, any line search method with Backtracking, or any Newton method that does not update the Hessian at minor iterations).

Is that a bug? I'm not sure it is. The user said there was a runtime/function evaluation limit, and that limit was reached. If they want the gradient they can always evaluate the function at that location. As a user, I'd much rather receive the location with a lower function value than have it thrown away.

vladimir-ch:
> The recorders don't need such signaling. They know EvaluationType and Location, so they can know which fields are valid. Similarly, the optimizers should know what's valid, since they are the ones requesting the data. The thought of having the NaN is just to have it to help prevent silent bugs from Methods using the Gradient when they previously requested a function. If you, as an implementer, see NaNs, you know you have a bug somewhere (assuming the function doesn't NaN).

Optimizers are ok, I was just trying to think of how Recorders could know that more than what EvaluationType indicates can be valid.

> Yes, this sounds reasonable. In the Convex world, I'm not sure this is true, but in the non-convex world (like this package), you probably always need at least the function value. I wish there were a way to not recompute the function value, but I would guess for most functions they need to compute the function value before they can compute the gradient (or at least have a significant function overlap).

We can put reducing the number of function interfaces on our thought stack.

> Is that a bug? I'm not sure it is. The user said there was a runtime/function evaluation limit, and that limit was reached. If they want the gradient they can always evaluate the function at that location. As a user, I'd much rather receive the location with a lower function value than have it thrown away.

I see what you mean. If we wanted to return a valid location, we would have to evaluate, which would violate user's limits. Sounds reasonable, but it has to be clearly stressed somewhere in the docs, it is not obvious. In any case, I didn't want to throw away the location with a lower function value, just somehow indicate that its fields may not be consistent. I guess that the return status indicates that, but as I said, it has to be documented, otherwise it is rather dangerous.

vladimir-ch:
Ok, I removed the proxies and instead I directly allocate any missing Location fields in evaluate(). I simplified invalidate() to NaN only the first element of Gradient and Hessian. I added the panic when NoEvaluation && different. And I added copying of the Hessian in copyLocation().
PTAL, @btracey

local.go
```diff
@@ -297,6 +307,18 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
 	return NotTerminated
 }
 
+func invalidate(loc *Location, f, grad, hess bool) {
```
btracey: Could you add a quick function comment? It's clear what the function does, so the comment just needs to say why.

local.go
```diff
@@ -320,44 +339,92 @@ func evaluate(funcInfo *functionInfo, evalType EvaluationType, xNext []float64,
...
 		if funcInfo.IsFunctionGradient {
+			if different {
+				invalidate(loc, false, false, true)
+			}
 			loc.F = funcInfo.functionGradient.FuncGrad(loc.X, loc.Gradient)
 			stats.FuncGradEvaluations++
 			return
 		}
+		if funcInfo.IsFunctionGradientHessian {
+			if loc.Hessian == nil {
```
btracey: I believe this is here, because the Hessian is only allocated if the method needs a hessian. Given this change, shouldn't IsFunctionGradient allocate a gradient if it's nil?

btracey: Nevermind. I understand.

local.go
```diff
@@ -320,44 +339,92 @@ func evaluate(funcInfo *functionInfo, evalType EvaluationType, xNext []float64,
...
-			loc.F = math.NaN()
-			for i := range loc.X {
-				loc.X[i] = math.NaN()
-			}
-			for i := range loc.Gradient {
-				loc.Gradient[i] = math.NaN()
-			}
+			// Optimizers should not request NoEvaluation at a new location.
+			// The intent and therefore an appropriate action are both unclear.
+			panic("no evaluation requested at new location")
```
btracey: Needs an "optimize: "

btracey:
LGTM with the minor fixes.

vladimir-ch:
Added a comment to invalidate() and optimize: to panic messages. Thanks for the comments, @btracey

PR #88, opened 2015-03-18
Add coordinate descent method.  Tweak HelicalValley test so that a step ...

btracey:
...of 1 from the initial condition does not have undefined function value

### DIFF
--- a/coordinatedescent.go
+++ b/coordinatedescent.go
@@ -1,4 +1,4 @@
-// Copyright ©2014 The gonum Authors. All rights reserved.
+// Copyright ©2015 The gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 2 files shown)


coordinatedescent.go
```diff
@@ -0,0 +1,91 @@
+package optimize
```
kortschak: header

btracey: Oops. Fixed.

vladimir-ch:
I think that the point of coordinate descent is that it is a gradient-free method, but your version does need the gradient. In such case I don't see much utility in this, because it needs the same info as gradient descent with much worse (none) convergence properties.

I had been thinking about coordinate descent before and I came to the conclusion that it would have to be implemented outside of `Linesearch` framework (which could be declared as derivative-based). Since we don't have the gradient, we don't know which direction is descent, so if one fails to find a significant decrease (after some maximum number of attempts), we try the opposite one and if that one fails too, we error. This would of course need a line search that does not use the Armijo/Wolfe conditions and another way of judging "significant decrease".

vladimir-ch:
In my edition of Nocedal&Wright there is a section about Pattern-search methods, of which coordinate descent is a special case. They also list an outline of an algorithm and give two examples of direction sets, coordinate direction set and one that they call simplex set. I would rather remove the need for the gradient and implement that algorithm instead.

btracey:
Makes sense that it's gradient free.

PR #90, opened 2015-03-24
Add initial Hessian to Settings

vladimir-ch:
* Added initial Hessian to Settings
* Changed how the initial location is constructed. The deep nesting in there is not nice and I'd like to get rid of it, but the idea is what matters to me at the moment: the initial function value is used always, but the gradient and Hessian only if they are not nil. If they are, evaluate() is called. If the user wanted to use Newton method on a solution obtained with other gradient-based method, this would be useful. The problem is that if the objective function does not implement methods that directly correspond to evaluation types, we could end up doing more computation than if we just ignored the initial data. We would be doing our best with what the user gave us, but the behaviour might still be slightly surprising and hard to document.

Seeing the code, I think I would revive the idea of putting StartTime into Stats and moving Runtime into Result.

PTAL, @btracey

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -82,7 +82,7 @@ type Method interface {
 
 	// Needs specifies information about the objective function needed by the
 	// optimizer beyond just the function value. The information is used
-	// internally for initialization and has to match evaluation types returned
+	// internally for initialization and must match evaluation types returned
 	// by Init() and Iterate() during the optimization process.
 	Needs() struct {
 		Gradient bool
--- a/local.go
+++ b/local.go
@@ -212,7 +212,7 @@ func getStartingLocation(funcInfo *functionInfo, method Method, initX []float64,
 		X: make([]float64, dim),
 	}
 	copy(loc.X, initX)
-	if method.Needs().Gradient {
+	if method.Needs().Gradient || method.Needs().Hessian {
 		loc.Gradient = make([]float64, dim)
 	}
 	if method.Needs().Hessian {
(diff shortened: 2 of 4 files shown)


btracey:
My original intent was that if the user wants initial information to be supplied, they supply all of the information needed at the initial location. The thought is that it reduces the magic a bit. The cost is not very much, for example, in the case you specify, the user just has to call the Hessian() method of their function. Feel free to disagree.

types.go
```diff
@@ -176,9 +176,10 @@ func (f functionInfo) satisfies(method Method) error {
 //
 // If Recorder is nil, no information will be recorded.
 type Settings struct {
-	UseInitialData       bool      // Use supplied information about the conditions at the initial x.
-	InitialFunctionValue float64   // Func(x) at the initial x.
-	InitialGradient      []float64 // Grad(x) at the initial x.
+	UseInitialData       bool            // Use supplied information about the conditions at the initial x.
+	InitialFunctionValue float64         // Func(x) at the initial x.
```
btracey: Can we remove the "Value" in InitialFunctionValue while we're changing this? The inconsistency is annoying.

btracey:
To be clear, the benefit is that it simplifies the rules. It's either an error or nothing is evaluated if settings.UseInitialValue = true, and everything is evaluated otherwise.

types.go
kortschak: It seems odd to call a float64 field be called Func. I understand the motivation, but I think InitialValue is a better name.

btracey: SGTM. It seemed weird to have a two word moniker for Function, but only a one word moniker for Gradient and Hessian.

kortschak: It would be one word, `InitialValue`.

btracey: Yes, understood.

vladimir-ch:
I agree, I only wanted to test the idea of not requiring the complete initial information just because we can. If the logic to evaluate it were simple, I would vote for keeping it, but it is not. I will update the PR to require the complete information and panic otherwise as it was before. I will also change InitialFunctionValue to InitialValue.

vladimir-ch:
PTAL, @btracey

vladimir-ch:
I have made one additional change that further simplifies the evaluation of the starting location (and will make our life easier in the future as well): if the method Needs().Hessian, then assume that it also Needs().Gradient. It is complicated, annoying and unrealistic to assume that a method that needs the Hessian doesn't need the gradient too. This should be documented somewhere, at Method.Needs()? Also, I wonder if we shouldn't panic when a method returns such Needs(). @kortschak @btracey

kortschak:
SGTM

If documented, then panicking is fine.

vladimir-ch:
Added panic() and a comment to Needs(). I hope the wording is ok.

kortschak:
That looks fine to me.

vladimir-ch:
Is this ok to merge, @btracey ?

local.go
```diff
@@ -59,6 +59,9 @@ func Local(f Function, initX []float64, settings *Settings, method Method) (*Res
 	if method == nil {
 		method = getDefaultMethod(funcInfo)
 	}
+	if method.Needs().Hessian && !method.Needs().Gradient {
```
btracey: I don't like this. If the method doesn't need the gradient, then it doesn't need the gradient. The method shouldn't have to know it will be used in Local. We can just change line 215 to add an  "|| method.Needs().Hessian".

interfaces.go
```diff
@@ -82,8 +82,9 @@ type Method interface {
 
 	// Needs specifies information about the objective function needed by the
 	// optimizer beyond just the function value. The information is used
-	// internally for initialization and has to match evaluation types returned
-	// by Init() and Iterate() during the optimization process.
+	// internally for initialization and must match evaluation types returned
+	// by Init() and Iterate() during the optimization process. If Hessian is
```
btracey: This last requirement doesn't need to exist. See other note.

btracey:
LGTM minus removing the panic. Sorry for not piping in earlier.

PR #91, opened 2015-03-26
Add neldermead

btracey:


btracey:
The main motivation is to add Nelder-Mead and some basic tests. I have also confirmed by hand that the implementation of Nelder-Mead matches that of the implementation in juliaopt for a few more cases (when the parameter values match). 

There is no way in the existing code to converge a gradient-free method, so add that capability. Also, rename some settings for clarity.

We'll have to think some more on the right set of tests to use for gradient-free methods, but this PR is already quite large.

local.go
```diff
@@ -256,12 +259,19 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
 	if iterType == MajorIteration || iterType == InitIteration {
 		if loc.Gradient != nil {
 			norm := floats.Norm(loc.Gradient, math.Inf(1))
-			if norm < settings.GradientAbsTol {
-				return GradientAbsoluteConvergence
+			if norm < settings.GradientThreshhold {
```
vladimir-ch: No double h

functionconvergence.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (fc *FunctionConvergence) Init(f float64) {
+	fc.absBest = f
+	fc.relBest = f
+	fc.absIter = 0
+	fc.relIter = 0
+}
+
+func (fc *FunctionConvergence) FunctionConverged(f float64) Status {
+	if f >= fc.absBest-fc.Absolute {
```
vladimir-ch: I would prefer this reorganized as in the relative test `fc.absBest - f <= fc.Absolute`.

neldermead.go
```diff
@@ -0,0 +1,292 @@
...
+	n.centroid = resize(n.centroid, dim)
+	n.reflectedPoint = resize(n.reflectedPoint, dim)
+
+	// Default parameter choices are chosen in a dimension-dependent way
+	// from http://www.webpages.uidaho.edu/~fuchang/res/ANMS.pdf
+	if n.Reflection == 0 {
+		n.Reflection = 1
+	}
+	if n.Expansion == 0 {
+		n.Expansion = 1 + 2.0/float64(dim)
```
vladimir-ch: Nitpicking: .0 here and below are not necessary

functionconvergence.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (fc *FunctionConvergence) Init(f float64) {
+	fc.absBest = f
+	fc.relBest = f
+	fc.absIter = 0
+	fc.relIter = 0
+}
+
+func (fc *FunctionConvergence) FunctionConverged(f float64) Status {
```
vladimir-ch: Just thinking aloud: What if the two tests were merged into one like
```go
if fc.best - f <= fc.Relative*math.Abs(f) + fc.Absolute {
    // ...
}
```
There would be only one return status `FunctionConvergence` (something would have to be renamed), but this would become simpler (not that it is complicated now).

neldermead.go
```diff
@@ -0,0 +1,292 @@
...
+		if n.fillIdx == dim {
+			// Successfully finished building initial simplex.
+			sort.Sort(nmVertexSorter{n.vertices, n.values})
+			computeCentroid(n.vertices, n.centroid)
+			return n.returnNext(nmReflected, xNext)
+		}
+		copy(xNext, loc.X)
+		xNext[n.fillIdx-1] -= 1
+		xNext[n.fillIdx] += 1
+		copy(n.vertices[n.fillIdx], xNext)
```
vladimir-ch: For consistency and keeping things together, it would be better to move this line to 157 and change `xNext` to `loc.X`.

vladimir-ch:
I'll look closer at the Nelder-Mead itself tomorrow.

vladimir-ch:
It seems to me that reusing `NelderMead` does not work properly, in the 4d test the two test functions give different minima although as far as I can tell they should not.

local.go
btracey: One of these days I'll learn to spell that word.

functionconvergence.go
btracey: It's harder to meet than either individually, but in most cases the net effect shouldn't be that large. If one is "much more important" than the contribution of the other is negligible, with the limiting case that the behavior is the same if one or the other is zero. The largest difference in behavior is when they're the similar order of magnitude, which is a factor of 2. This doesn't that different. Thus, the user can still think of the convergence parameters individually when choosing values, and if they really care this rule is much clearer than the currently implemented rule. 

So, SGTM.

neldermead.go
btracey: I'm moving it, but this code right now assumes specific values in xNext, which is bad, so I'm also fixing that.

btracey:
Added a test for reusing the gradient-free method. Apparently that's not the problem, something weirder is happening. Looking into it.

btracey:
Ah! Got it.

btracey:
Fixed. PTAL.

types.go
```diff
@@ -239,9 +247,13 @@ type Settings struct {
...
 func DefaultSettings() *Settings {
 	return &Settings{
-		GradientAbsTol: defaultGradientAbsTol,
-		FunctionAbsTol: math.Inf(-1),
-		Recorder:       NewPrinter(),
+		GradientThreshold: defaultGradientAbsTol,
+		FunctionThreshold: math.Inf(-1),
+		Recorder:          NewPrinter(),
+		FunctionConverge: &FunctionConverge{
+			Absolute:   1e-10,
```
vladimir-ch: Can we set also Relative to non-zero value? Magnitude of the minimum is unknown, if it is zero, Absolute will matter, if it is non-zero, Relative will matter. Perhaps 1e-6?

neldermead.go
```diff
@@ -0,0 +1,307 @@
...
+// algorithm described in
+//
+//  http://epubs.siam.org/doi/pdf/10.1137/S1052623496303470
+//
+// If an initial simplex is provided, it is used and initLoc is ignored. If
+// InitialVertices and InitialValues are both nil, an initial simplex will be
+// generated automatically using the initial location as one vertex, and each
+// additional vertex as SimplexSize away in one dimension.
+//
+// If the simplex update parameters (Reflection, etc.)
```
vladimir-ch: Re-wrap this paragraph?

neldermead.go
```diff
@@ -0,0 +1,307 @@
...
+//
+//  http://www.webpages.uidaho.edu/~fuchang/res/ANMS.pdf
+type NelderMead struct {
+	InitialVertices [][]float64
+	InitialValues   []float64
+	Reflection      float64 // Reflection parameter (>0)
+	Expansion       float64 // Expansion parameter (>1)
+	Contraction     float64 // Contraction parameter (>0, <1)
+	Shrink          float64 // Shrink parameter (>0, <1)
+	SimplexSize     float64 // size of auto-constructed initial simplex
```
vladimir-ch: Capital S in the comment?

neldermead.go
```diff
@@ -0,0 +1,307 @@
...
+
+	reflection  float64
+	expansion   float64
+	contraction float64
+	shrink      float64
+
+	vertices [][]float64 // location of the vertices sorted in ascending f
```
vladimir-ch: Capital letters also here and below?

neldermead.go
```diff
@@ -0,0 +1,307 @@
...
+	n.values[dim] = loc.F
+	n.fillIdx = 0
+	copy(xNext, loc.X)
+	xNext[0] += n.SimplexSize
+	n.lastIter = nmInitialize
+	return FuncEvaluation, InitIteration, nil
+}
+
+// computeCentroid computes the centroid of all the simplex vertices except the
+// final one
```
vladimir-ch: Dot?

neldermead.go
```diff
@@ -0,0 +1,307 @@
...
+		sort.Sort(nmVertexSorter{n.vertices, n.values})
+		computeCentroid(n.vertices, n.centroid)
+		return n.returnNext(nmReflected, xNext)
+	default:
+		panic("unreachable")
+	}
+}
+
+// returnNext finds the next location to evaluate, stores the location in xNext,
+// and returns the data
```
vladimir-ch: Dot?

vladimir-ch:
And also, consider moving `FunctionConverge` to `types.go`?

LGTM with the above nitpicks and minor changes. Great work, we have a derivative-free optimization!

However ... I'd like you to clean up the commits in this PR ... it will be slightly complicated. I tried to come up with the easiest way possible, perhaps there is a cleverer way, but this is how I would do it (the merge commit from master to this branch complicates the things slightly):

```sh
$ git checkout add-neldermead
# make sure that everything is committed and the working tree is clean
$ git pull     # just in case
$ git rebase -i ec4055c
```

Now in the editor you have to delete all the lines that belong to commits that were included by the merge from master. They are all my commits, consecutive, 14 of them, so not so difficult to identify. After you have made the above nitpicks and minor changes, there will be N commits in this PR (now 11), after removing the line with my commits you should have N-1 lines for rebase (one commit was the merge commit that does not appear in the editor). Save and exit. Then:

```sh
$ git rebase master
```

Everything should go smoothly without conflicts and you should end up with a branch that contains only your new commits. Now I think the best way is to get rid of these commits while keeping the changes and create three new commits, one for FunctionConverge, one for NelderMead and one for tests:

```sh
$ git reset HEAD~N-1    # replace N-1 with the actual value from above!
```

Now you have add-neldermead branch without any commits, but with unstaged changes and untracked files. From now on it is just routine git, so for example for the first commit:

```sh
$ git add functionconvergence.go  # if you decide to keep it after all
$ git add -p
# follow instructions to stage pieces of code that go with FunctionConverge
$ git commit -m "Add FunctionConverge"
```

Repeat the same for NelderMead and its tests, you will end up with three _beautiful_ commits. Now on to merging:

```sh
$ git push -f   # to update the branch on GitHub and keep a connection to this PR
$ git checkout master
$ git merge --no-ff add-neldermead
# no fast-forward, I think it is a good idea to keep these three commits
# together, I have been doing it recently as well. Edit the commit message
# and include "Closes #18". Save and exit.
$ git push
```

That's it, the issue and PR are both automatically closed, so just delete add-neldermead branch on github. Then `git fetch -p` and `git branch -d add-neldermead`  to delete it also locally. Finished. Am I not too verbose?

neldermead.go
btracey: Done.

neldermead.go
btracey: Done.

neldermead.go
btracey: Done.

neldermead.go
btracey: Done.

types.go
btracey: I'd like the default to be "We really found the optimum". I set it to 1e-8.

PR #92, opened 2015-03-27
ORable EvaluationTypes; required Gradient and/or Hessian interfaces

vladimir-ch:
A quick, experimental implementation that deals with the problem of too many combinations of what method needs, what function can evaluate, what should be the next evaluation given the previous evaluation, etc.

* Gradient and Hessian interfaces are required if Method.Needs() indicates so
* FunctionGradient and FunctionGradientHessian are optional
* EvaluationTypes are ORable and any combination is acceptable
* loc.Gradient does not need to be allocated if method needs Hessian
* evaluate() is greatly simplified. If we removed the combined, optional interfaces completely, it could be even simpler.
* Adding constraints evaluation in the future would be straightforward

PTAL, @btracey

### DIFF
--- a/linesearch.go
+++ b/linesearch.go
@@ -55,8 +55,8 @@ func (ls *Linesearch) Init(loc *Location, f *FunctionInfo, xNext []float64) (Eva
 
 func (ls *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
 	if ls.iterType == SubIteration {
-		// We needed to evaluate the gradient, so now we have it and can
-		// announce MajorIteration.
+		// We needed to evaluate invalid fields of Location. Now we have them
+		// and can announce MajorIteration.
 		ls.iterType = MajorIteration
 		copy(xNext, loc.X)
 		return NoEvaluation, ls.iterType, nil
@@ -76,19 +76,17 @@ func (ls *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, I
 		Derivative: projGrad,
 	}
 	if ls.Method.Finished(lsLoc) {
-		if ls.lastEvalType == FuncEvaluation && loc.Gradient != nil {
-			// We have the function value at the current location, but we don't
-			// have the gradient, so get it before announcing MajorIteration.
-			ls.iterType = SubIteration
-			copy(xNext, loc.X)
-			return GradEvaluation, ls.iterType, nil
-		}
-		// The linesearch is finished. Announce so with an update to
-		// MajorIteration. The function value and gradient is already known, so
-		// no function evaluations are necessary.
-		ls.iterType = MajorIteration
 		copy(xNext, loc.X)
-		return NoEvaluation, ls.iterType, nil
+		// Check if the last evaluation evaluated all fields of Location.
+		complEval := complementEval(loc, ls.lastEvalType)
+		if complEval == NoEvaluation {
+			// Location is complete and MajorIteration can be announced directly.
+			ls.iterType = MajorIteration
+			return complEval, ls.iterType, nil
+		}
+		// Location is not complete, evaluate its invalid fields in SubIteration.
+		ls.iterType = SubIteration
+		return complEval, ls.iterType, nil
 	}
 
 	// Line search not done, just iterate.
--- a/local.go
+++ b/local.go
@@ -354,8 +354,7 @@ func invalidate(loc *Location) {
 // loc are set to NaN.
 // evaluate panics if the function does not support the requested evalType.
 func evaluate(f *functionInfo, evalType EvaluationType, xNext []float64, loc *Location, stats *Stats) {
-	different := !floats.Equal(loc.X, xNext)
-	if different {
+	if !floats.Equal(loc.X, xNext) {
 		if evalType == NoEvaluation {
 			// Optimizers should not request NoEvaluation at a new location.
 			// The intent and therefore an appropriate action are both unclear.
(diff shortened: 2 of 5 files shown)


types.go
```diff
@@ -15,33 +15,34 @@ import (
...
 type EvaluationType int
 
 const (
-	NoEvaluation EvaluationType = iota
-	FuncEvaluation
-	GradEvaluation
-	HessEvaluation
-	FuncGradEvaluation
-	FuncGradHessEvaluation
+	NoEvaluation           EvaluationType = 0
```
btracey: I'm bad with bit math (hopefully this PR will help), so pardon my ignorance.

I would like it to be that FuncGradEvaluation exists implicitly, and instead Method returns 
````
FuncEvaluation + GradEvaluation
````
to express that it wants both. This is very friendly from an implementation perspective. 

````
FuncEvaluation + GradEvaluation + ConstraintEvaluation
````
We have 4 bits (Func, grad, hessian, constraint), and so we can test that result by looking at the bit (something involving bit shift?).

Comment (btracey) [inline types.go]: Ah, I understand the bitwise-and now. Still, we should be able to return using "+" rather than using the bitwise or, right?

Comment (kortschak) [inline types.go]: Don't write it with `+`. If this is done and you typo a repeated flag, then you get a shift.

Also, @vladimir-ch, you can use `1<<iota` for the second line and elide the next two.

**Comment by vladimir-ch** [comment]
When you have a chance, PTAL @btracey

File: linesearch.go [hunk 2]
```diff
@@ -76,19 +76,17 @@ func (ls *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, I
 		Derivative: projGrad,
 	}
 	if ls.Method.Finished(lsLoc) {
-		if ls.lastEvalType == FuncEvaluation && loc.Gradient != nil {
-			// We have the function value at the current location, but we don't
-			// have the gradient, so get it before announcing MajorIteration.
+		// Check if the last evaluation evaluated all fields of Location.
+		complEval := complementEval(loc, ls.lastEvalType)
```
btracey: I think you can reduce the branching here by moving the copy before the complEval check and then putting the return inside. It's a simple if statement, but the consequences are different (finish an iteration vs. don't finish an iteration), so multiple return points feels clearer to me.

local.go
```diff
@@ -338,130 +338,52 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
...
+	if !floats.Equal(loc.X, xNext) {
+		if evalType == NoEvaluation {
 			// Optimizers should not request NoEvaluation at a new location.
 			// The intent and therefore an appropriate action are both unclear.
 			panic("optimize: no evaluation requested at new location")
 		}
-		return
-	default:
+		copy(loc.X, xNext)
+		invalidate(loc, evalType)
```
btracey: I think it's clearer if invalidate just corrupts the whole structure (without any logic). This way:
1) Give a new x, the location fields are immediately unusable. They get usable based on the eval type.

local.go
```diff
@@ -338,130 +338,52 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
...
+		stats.GradEvaluations++
+		evalType &= ^GradEvaluation
+	}
+	if evalType&HessEvaluation != 0 {
+		f.hessian.Hess(loc.X, loc.Hessian)
+		stats.HessEvaluations++
+		evalType &= ^HessEvaluation
+	}
+
+	if evalType != NoEvaluation {
```
btracey: I think this is a good way to check for bad evalTypes, but the error message should return the original evalType specified by the optimizer. Probably something like
````
toEval := evalType
if evalType&FuncEvaluation != 0 {
     toEval &= ^FuncEvaluation
} 
...
if toEval != NoEvaluation {
     panic(fmt.Sprintf("optimize: unknown evaluation type %v", evalType))   
}
````

Now that evalType is just a number, do we really need fmt?

File: types.go [hunk 5]
```diff
@@ -15,33 +15,39 @@ import (
 
 const defaultGradientAbsTol = 1e-6
 
-// EvaluationType is used by the optimizer to specify information needed
-// from the objective function.
-type EvaluationType int
+// EvaluationType is used by the optimizer to specify information needed from
+// the objective function. The types can be composed together using the binary
+// or operator.
```
btracey: Our users probably haven't used binary or much, so "using the binary or operator, for example 'FunctionEvaluation | GradEvaluation' to evaluate both the function value and the gradient"

types.go
```diff
@@ -15,33 +15,39 @@ import (
...
-	FuncEvaluation
+	NoEvaluation   EvaluationType = 0
+	FuncEvaluation EvaluationType = 1 << iota
 	GradEvaluation
 	HessEvaluation
-	FuncGradEvaluation
-	FuncGradHessEvaluation
+)
+
+// Combined evaluation types.
```
btracey: This should stay in this CL to keep the changes small, but I think we should remove these. It's easy to write in the optimizers, will make it clearer to new users how to compose the evaluations (by seeing the example in our code) and keeps the API surface smaller.

types.go
```diff
@@ -15,33 +15,39 @@ import (
...
 }
 
-var evaluationStrings = [...]string{
-	"NoEvaluation",
-	"FuncEvaluation",
-	"GradEvaluation",
-	"HessEvaluation",
-	"FuncGradEvaluation",
-	"FuncGradHessEvaluation",
+var evaluationStrings = map[EvaluationType]string{
```
btracey: I'm pretty sure these are worthless now, since mostly the constants will be composed together.

linesearch.go
vladimir-ch: Done

local.go
vladimir-ch: Done

local.go
vladimir-ch: My intent was to return really only the unknown part of evalType, but you are right that returning the original evalType is better, it may make the tracing of its origin easier. Anyway, done.

types.go
vladimir-ch: Done

types.go
vladimir-ch: Agreed. I will do it when this is merged.

types.go
vladimir-ch: We can remove `evaluationStrings` but I would like to keep `EvaluationType.String()` anyway. Evaluation types are just numbers, but the bits can be represented as strings in the same way as they appear in the code. On the other hand, returning a string "EvaluationType(7)" is one line of code, returning "FuncEvaluation|GradEvaluation|HessEvaluation" will be much longer unless we keep evaluationStrings and put all combinations in it.

kortschak: Agree. There should be some way of representing the values that is more meaningful than 7 (which is how fmt would present it).

btracey: Would you say that "EvaluationType(7)" is more meaningful? It's likely we'll only have one more EvaluationType, which is NonlinearConstraints (other constraints are fixed, and at the moment I'm imagining that linear and non-linear are part of one function call). This is 16 combinations which is enumerable if we wanted to keep evaluationStrings. We could also return something like
````
EnumerationType(V: 11, F: true, G: true, H: false, C: true, E: 0)
EnumerationType(V: 43, F: true, G: true, H: false, C: true, E: 32)
````
where "E" is extra and would help users who get an "unknown evaluation type" panic.

File: local.go [hunk 8]
```diff
@@ -334,134 +334,56 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
...
-func invalidate(loc *Location, f, grad, hess bool) {
-	if f {
-		loc.F = math.NaN()
-	}
-	if grad && loc.Gradient != nil {
+// invalidate marks all fields of Location with NaNs and it is the
+// responsibility of Method to issue such EvaluationTypes that make Location
+// valid again. Invalidation is done internally as a help for implementers to
+// detect silent bugs in Methods using inconsistent Location, e.g., using
+// Gradient after FuncEvaluation request.
```
btracey: using Gradient without a GradEvaluation request.

types.go
```diff
@@ -15,33 +15,40 @@ import (
 
 const defaultGradientAbsTol = 1e-6
 
-// EvaluationType is used by the optimizer to specify information needed
-// from the objective function.
-type EvaluationType int
+// EvaluationType is used by the optimizer to specify information needed from
```
btracey: EvaluationType is used by a Method to specify the objective-function information needed at an x location. The types...

types.go
btracey: But if we're just using EvaluationType(7), which I think is okay, then evalutionStrings are unused and can be removed.

btracey:
LGTM with comment fixed addressed. I don't think decision on evaluation strings needs to happen here.

vladimir-ch:
Fixed both comments, merging. Removal of combined evaluation types and decision on evaluation strings deferred to the next PR.

PR #95, opened 2015-04-14
Bisection: Change procedure to work toward the global minimum of the lin...

btracey:
...esearch, not any local minimum.

The previous update rules could find a point of low function value, but stop the linesearch at a higher function value. This PR changes the behavior to have the linesearch stop at a point that has at least as low of a function value as the lowest returned point (see the discussion in issue 93). The point returned still meets the Wolfe conditions as originally specified. This should help stay in lower regions of attraction.

### DIFF
--- a/bisection.go
+++ b/bisection.go
@@ -57,7 +57,20 @@ func (b *Bisection) Init(loc LinesearchLocation, step float64, _ *FunctionInfo) 
 }
 
 func (b *Bisection) Finished(loc LinesearchLocation) bool {
-	return StrongWolfeConditionsMet(loc.F, loc.Derivative, b.initF, b.initGrad, b.currStep, 0, b.GradConst)
+	// Don't finish the linesearch until a minimum is found that is better than
+	// the best point found so far. We want to end up in the lowest basin of
+	// attraction
+	minF := b.initF
+	if b.maxF < b.initF {
+		minF = b.maxF
+	}
+	if b.minF < minF {
+		minF = b.minF
+	}
+	if StrongWolfeConditionsMet(loc.F, loc.Derivative, minF, b.initGrad, b.currStep, 0, b.GradConst) {
+		return true
+	}
+	return false
 }
 
 func (b *Bisection) Iterate(loc LinesearchLocation) (float64, EvaluationType, error) {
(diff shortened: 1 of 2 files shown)


btracey:
@vladimir-ch sorry, this fails the CG tests. Some of them just need to have the tolerances loosened a bit, but the Wood test fails for a bunch of the CG step sizers even with a tolerance of 1e-2. What about CG would make this less robust?

bisection.go
```diff
@@ -57,7 +57,20 @@ func (b *Bisection) Init(loc LinesearchLocation, step float64, _ *FunctionInfo)
...
+	// the best point found so far. We want to end up in the lowest basin of
+	// attraction
+	minF := b.initF
+	if b.maxF < b.initF {
+		minF = b.maxF
+	}
+	if b.minF < minF {
+		minF = b.minF
+	}
+	if StrongWolfeConditionsMet(loc.F, loc.Derivative, minF, b.initGrad, b.currStep, 0, b.GradConst) {
```
vladimir-ch: I may not see through your intention here, but replacing initF with the minimum of initF, minF and maxF but keeping initGrad seems to me like a possible cause of trouble. For Armijo condition, you are essentially asking for initGrad-proportional decrease from minF value, which might be impossible to achieve, especially close to a minimum and/or for difficult functions. This problem might be more prominent with CG, because the directions are not well-scaled and not so good (just guessing here, I haven't go around to test it). Am I missing something?

btracey: The problem that I was seeing was this:
````
\                  /
   \    /\       /
     \/   \    /
            \/             
````
The line search starts on the left, steps to a point on the right. This point on the right has a lower function value than the leftmost minimum, but has a high gradient and so does not meet the Wolfe conditions. The linesearch bisects, and ends up in the left minimum. The gradient is small, but the function value is higher than we had previously found. The goal of this change is to fix the behavior of Bisection such that the linesearch terminates in the rightmost minimum instead of the leftmost.

The Wolfe conditions are that there is a significant decrease in the function value (which for Bisection means that f_final <= f_start), and that there is a proportional decrease in the gradient |g_final| < alpha |g_start|. One must do the gradient comparison with reference to g_init, otherwise the Wolfe condition may not be met. The function criterion is that f_final <= f_start, but this is f_final <= f_min <= f_start, so this termination criteria still meets Wolfe.

Comment (btracey) [inline bisection.go]: To add, in particular, by testing with min_f rather than init_f, when we find the minimum on the left, we don't want to converge -- we want to find the minimum on the right. The linesearch knows this minimum exists because we have something like
````
x_init = 0, f_init = 10, g_init = -1
x_left = 1, f_left = 8, g_left = 0.1
x_right = 2, f_right = 5, g_right = 3
````
so there must be a local minimum and maximum in order to get from x_left to x_right. When we see the function value is higher at x_left, instead of stopping because it's the first point we've found that meets Wolfe, we decide to find that local minimum we know must exist near x_right, and so we choose to bisect between x_left and x_right.

Comment (vladimir-ch) [inline bisection.go]: I always forget that you use 0 in the Armijo condition. Give me time to look into this.

Comment (vladimir-ch) [inline bisection.go]: Mea maxima culpa, Wood.Grad() was incorrect. Have a look at #97 and retest after merging, @btracey

Comment (vladimir-ch) [inline bisection.go]: Merged. On `fix-bisection` branch, please, `git pull` and `git rebase origin/master`.

**Comment by btracey** [comment]
Okay, now it passes the tests. I think this lowers the tolerance just because of floating point noise. Since we now need to be lower than the lowest point, if the optimizer finds the random point where the value was rounded down instead of up, it can be flaky.

````
function value:      -------------_--------------_----------------
gradient value:      \                                      /
                                      \                         /
                                             \               /
                                                       \/
````
If the linesearch hits either of those blips, it will never find a gradient sufficiently small. Note that this problem exists already, namely if MajorIteration begins at one of the blips. I feel like finding a better minimum is a reasonable price to pay for not being able to optimize quite as far. This problem can be fixed by treating function values within a certain ULP as being the same. We had this in before, but it was causing other issues so it was removed.

**Comment by vladimir-ch** [comment]
LGTM

Before merging, try to bring `fix-bisection` branch up-to-date by rebasing it on top of master instead of merging master into it. Something like this should work (being on `fix-bisection` branch):
```
$ git fetch
$ git rebase origin/master
$ git push -f
```
Then you can merge the pull request as usual using the web ui.

PR #98, opened 2015-04-20
Remove combined evaluation types

vladimir-ch:
PTAL, @btracey 

In the end I decided to format EvaluationType using your suggestion. Using just `EvaluationType(number)` would be also fine, because non-implementers of outside Methods would almost never get to see evaluation types default formatted, only possibly via Recorders. On the other hand, implementers of Methods could get to see a default-formatted evaluation type in a panic, in which case a better formatted information could be handy.

### DIFF
--- a/local.go
+++ b/local.go
@@ -243,13 +243,12 @@ func getStartingLocation(funcInfo *functionInfo, method Method, initX []float64,
 			loc.Hessian.CopySym(initH)
 		}
 	} else {
-		switch {
-		case loc.Hessian != nil:
-			evalType = FuncGradHessEvaluation
-		case loc.Gradient != nil:
-			evalType = FuncGradEvaluation
-		default:
-			evalType = FuncEvaluation
+		evalType = FuncEvaluation
+		if loc.Gradient != nil {
+			evalType |= GradEvaluation
+		}
+		if loc.Hessian != nil {
+			evalType |= HessEvaluation
 		}
 		evaluate(funcInfo, evalType, loc.X, loc, stats)
 	}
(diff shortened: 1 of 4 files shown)


local.go
```diff
@@ -299,22 +298,19 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
 	}
 
 	if settings.FuncEvaluations > 0 {
-		totalFun := stats.FuncEvaluations + stats.FuncGradEvaluations + stats.FuncGradHessEvaluations
-		if totalFun >= settings.FuncEvaluations {
+		if stats.FuncEvaluations >= settings.FuncEvaluations {
```
btracey: This is the best part of the simplification.

btracey:
LGTM when you and Dan decide on the correct representation.

local.go
vladimir-ch: I know :-) Closing #9 automatically from 09af8f9 gave me much delight.

vladimir-ch:
I will keep `0b%b`. Using octal is a nice idea, but the number of evaluation types will surely increase, at least by one. We can revisit this later, keeping binary or changing to hexadecimal. Do you agree, @kortschak ?

kortschak:
Yeah, I'm fine with that. At the moment, the number of cases is likely to be small, so binary trumps octal in this case.

BTW I'm wondering where those comments went? That previous change doesn't show up here as being even an outdated diff.

LGTM

vladimir-ch:
Indeed. The formatting already looked unusual when I posted my previous comment, but the comments were there, now they are gone. I didn't do anything unusual with the commits, just standard rebase and push -f. But I noticed an icon I had never seen before, next to a commit that had line comments, so GitHub might be introducing some new features/ui elements ...

kortschak:
>GitHub might be introducing some new features/ui elements

The random comment loss feature. Excellent. At least it was not something that mattered.

PR #99, opened 2015-04-20
Add Newton method

vladimir-ch:
Has the time really come for Newton?

An implementation of a pure, locally convergent Newton method without linesearch and of a modified, globally convergent Newton method with a linesearch.

The pure Newton is rather simple now that mat64 has symmetric matrices and Cholesky decomposition.

The modification in ModifiedNewton is a simple one, just add multiples of identity to the Hessian until it becomes positive definite. I did not venture into the modified Cholesky method at the moment but I can if there is a desire to have it.

I added three functions for resizing/reusing matrices in a similar way we already resize/reuse slices. Experts on matrix package @kortschak @btracey might have a better suggestion.

No tests at the moment, first we need to add Hess() method to functions, then modify the testing itself to accomodate Hessian-based methods.

PTAL, @btracey

### DIFF
--- a/newton.go
+++ b/newton.go
@@ -5,120 +5,59 @@
 package optimize
 
 import (
-	"errors"
 	"math"
 
 	"github.com/gonum/floats"
 	"github.com/gonum/matrix/mat64"
 )
 
-// Newton implements pure Newton's method without linesearch for unconstrained
-// minimization of functions that satisfy Gradient and Hessian interfaces.
+// Newton implements a modified Newton's method for Hessian-based unconstrained
+// minimization. It applies regularization when the Hessian is not positive
+// definite, and it can converge to a local minimum from any starting point.
 //
-// Newton's method iteratively forms a quadratic model to the objective
-// function f and tries to minimize this approximate model. It generates a
-// sequence of locations x_k by means of
+// Newton iteratively forms a quadratic model to the objective function f and
+// tries to minimize this approximate model. It generates a sequence of
+// locations x_k by means of
 //  solve H_k d_k = -∇f_k for d_k,
-//  x_{k+1} = x_k + d_k,
-// where H_k is the Hessian matrix of f at x_k.
+//  x_{k+1} = x_k + α_k d_k,
+// where H_k is the Hessian matrix of f at x_k and α_k is a step size found by
+// a line search.
 //
-// Note that Newton's method is only locally convergent. This means that for
-// starting points close to a minimizer the sequence x_k converges rapidly with
-// a quadratic rate. However, away from a minimizer the quadratic approximation
-// may not be valid, the Hessian may not be positive definite and d_k may not
-// be a descent direction. In such case Newton fails with an error.
-//
-// For a globally convergent Hessian-based method, see ModifiedNewton. If the
-// Hessian matrix cannot be formed explicitly or if the computational cost of
-// its factorization is prohibitive, BFGS or L-BFGS methods can be used
-// instead.
-type Newton struct {
-	chol *mat64.TriDense
-	step *mat64.Dense
-}
-
-func (n *Newton) Init(loc *Location, f *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
-	dim := len(loc.X)
-	n.chol = resizeTriDense(n.chol, dim)
-	r, _ := n.step.Dims()
-	if dim < r {
-		n.step = n.step.View(0, 0, dim, 1).(*mat64.Dense)
-	} else if dim > r {
-		n.step.Grow(dim, 1)
-	}
-	return n.computeNextX(loc, xNext)
-}
-
-func (n *Newton) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
-	return n.computeNextX(loc, xNext)
-}
-
-func (n *Newton) computeNextX(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
-	pd := n.chol.Cholesky(loc.Hessian, false)
-	if !pd {
-		return NoEvaluation, NoIteration, errors.New("optimize: Hessian is not positive definite")
-	}
-	n.step.SolveCholesky(n.chol, mat64.NewDense(len(loc.Gradient), 1, loc.Gradient))
-	floats.AddScaledTo(xNext, loc.X, -1, n.step.RawMatrix().Data)
-	return FuncEvaluation | GradEvaluation | HessEvaluation, MajorIteration, nil
-}
-
-func (n *Newton) Needs() struct {
-	Gradient bool
-	Hessian  bool
-} {
-	return struct {
-		Gradient bool
-		Hessian  bool
-	}{true, true}
-}
-
-// ModifiedNewton implements a modified Newton's method for unconstrained
-// minimization of functions that satisfy Gradient and Hessian interfaces. It
-// is a globally convergent line-search method that can be applied from any
-// starting point.
-//
-// ModifiedNewton method is based on the Newton step which computes the
-// line-search direction d_k as the solution of the linear system
-//  H_k d_k = -∇f_k,
-// where H_k is the Hessian matrix of the objective function. Away from a
-// minimizer H_k may not be positive definite and d_k may not be a descent
-// direction. ModifiedNewton implements a Hessian modification strategy that
-// tries to add successively larger multiples of identity to H_k until it
-// becomes positive definite. Note that the repeated trial factorization of the
+// Away from a minimizer H_k may not be positive definite and d_k may not be a
+// descent direction. Newton implements a Hessian modification strategy that
+// adds successively larger multiples of identity to H_k until it becomes
+// positive definite. Note that the repeated trial factorization of the
 // modified Hessian involved in this process can be computationally expensive.
-// For points in the vicinity of a minimizer ModifiedNewton reduces to a pure
-// Newton's method with its quadratic convergence rate.
 //
 // If the Hessian matrix cannot be formed explicitly or if the computational
 // cost of its factorization is prohibitive, BFGS or L-BFGS quasi-Newton method
 // can be used instead.
-type ModifiedNewton struct {
+type Newton struct {
 	// LinesearchMethod is a method used for selecting suitable steps along the
-	// descent direction d. Steps should satisfy the Wolfe, Goldstein or Armijo
-	// conditions. If LinesearchMethod == nil, an appropriate default is
-	// chosen.
+	// descent direction d. Steps should satisfy at least one of the Wolfe,
+	// Goldstein or Armijo conditions. If LinesearchMethod == nil, an
+	// appropriate default is chosen.
 	LinesearchMethod LinesearchMethod
-	// Increase is a factor by which a scalar tau is successively increased so
-	// that (H + tau*I) is positive definite. Larger values reduce the number
-	// of trial Hessian factorizations, but also reduce the second-order
+	// Increase is the factor by which a scalar tau is successively increased
+	// so that (H + tau*I) is positive definite. Larger values reduce the
+	// number of trial Hessian factorizations, but also reduce the second-order
 	// information in H.
 	// Increase must be greater than 1. If Increase is 0, it is defaulted to 5.
 	Increase float64
 
 	linesearch *Linesearch
 
-	hess *mat64.SymDense
-	chol *mat64.TriDense
+	hess *mat64.SymDense // Storage for a copy of the Hessian matrix.
+	chol *mat64.TriDense // Storage for the Cholesky factorization.
 	tau  float64
 }
 
-func (n *ModifiedNewton) Init(loc *Location, f *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
+func (n *Newton) Init(loc *Location, f *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
 	if n.Increase == 0 {
 		n.Increase = 5
 	}
 	if n.Increase <= 1 {
-		panic("optimize: ModifiedNewton.Increase must be greater than 1")
+		panic("optimize: Newton.Increase must be greater than 1")
 	}
 	if n.LinesearchMethod == nil {
 		n.LinesearchMethod = &Bisection{}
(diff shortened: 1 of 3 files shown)


types.go
```diff
@@ -257,3 +257,24 @@ func resize(x []float64, dim int) []float64 {
 	}
 	return x[:dim]
 }
+
+func resizeDense(m *mat64.Dense, r, c int) *mat64.Dense {
```
kortschak: Use Dense.View for this.

The other types don't have that though, so leave them until we support that.

Note @btracey, this use of View is not supported as well as it used to be because View returns a Matrix(*Dense) and this could all be done in place.

newton.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+		// Try to apply the Cholesky decomposition.
+		pd := n.chol.Cholesky(n.hess, false)
+		if pd {
+			break
+		}
+		// Modified Hessian is not PD, so increase tau.
+		n.tau = math.Max(n.Increase*n.tau, 0.001)
+	}
+	d := mat64.NewDense(dim, 1, dir)
```
kortschak: Add a comment that this is operating via side effect on dir's backing array.

newton.go
```diff
@@ -0,0 +1,203 @@
...
+type Newton struct {
+	chol *mat64.TriDense
+	step *mat64.Dense
+}
+
+func (n *Newton) Init(loc *Location, f *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
+	dim := len(loc.X)
+	n.chol = resizeTriDense(n.chol, dim)
+	r, _ := n.step.Dims()
+	if dim < r {
```
vladimir-ch: Do you mean using Grow and View like this, @kortschak ?

kortschak: Yes, that looks good.

newton.go
```diff
@@ -0,0 +1,203 @@
...
+// Note that Newton's method is only locally convergent. This means that for
+// starting points close to a minimizer the sequence x_k converges rapidly with
+// a quadratic rate. However, away from a minimizer the quadratic approximation
+// may not be valid, the Hessian may not be positive definite and d_k may not
+// be a descent direction. In such case Newton fails with an error.
+//
+// For a globally convergent Hessian-based method, see ModifiedNewton. If the
+// Hessian matrix cannot be formed explicitly or if the computational cost of
+// its factorization is prohibitive, BFGS or L-BFGS methods can be used
+// instead.
```
btracey: I'm not sure this should be in this package. The package is for general nonlinear minimization, and it is very easy for a general nonlinear function to have a non-positive-definite hessian, or to diverge in the iterative Newton procedure. This algorithm is useful somewhere, but I would rather have the Newton method in this package be one that can handle the general class of functions.

newton.go
```diff
@@ -0,0 +1,203 @@
...
+} {
+	return struct {
+		Gradient bool
+		Hessian  bool
+	}{true, true}
+}
+
+// ModifiedNewton implements a modified Newton's method for unconstrained
+// minimization of functions that satisfy Gradient and Hessian interfaces. It
+// is a globally convergent line-search method that can be applied from any
```
btracey: We need to find an alternate expression to 'globally convergent'. That implies it will find a global minimum.

// Newton implements a modified Newton's method for Hessian-based 
// unconstrained minimization. It applies regularization when the 
// Hessian is non-positive-definite, and uses line-searches so it can
// converge to a local minimum from any starting point.

newton.go
```diff
@@ -0,0 +1,203 @@
...
+				n.hess.SetSym(i, i, loc.Hessian.At(i, i)+n.tau)
+			}
+		}
+		// Try to apply the Cholesky decomposition.
+		pd := n.chol.Cholesky(n.hess, false)
+		if pd {
+			break
+		}
+		// Modified Hessian is not PD, so increase tau.
+		n.tau = math.Max(n.Increase*n.tau, 0.001)
```
btracey: Why is there a maximum here?

newton.go
```diff
@@ -0,0 +1,203 @@
...
+		if a < minA {
+			minA = a
+		}
+	}
+	if minA > 0 {
+		// All diagonal elements are positive, so try to factorize non-modified
+		// Hessian first.
+		n.tau = 0
+	} else if n.tau == 0 {
+		// Use the last tau that gave PD modified Hessian.
```
btracey: I don't understand the logic. If all the elements are positive, try the normal solve. Agreed. If they aren't all positive, only modify tau if the last solve went fine? What about the case where minA <= 0 and n.tau != 0?

newton.go
btracey: This doesn't actually increase tau if n.tau == 0

newton.go
vladimir-ch: I wasn't sure either. I included it because the method is very simple, just the Hessian and its factorization, no line search, and the docs say that it can fail if the starting point is far from a minimum. I have no problem removing it.

newton.go
vladimir-ch: I think that "globally convergent method" is a standard terminology for a method that converges from any starting point, not that it can find a global minimum. I agree that it can be confusing, though. It is probably best to avoid this term.

newton.go
vladimir-ch: If n.tau == 0, it does increase to 0.001.

newton.go
vladimir-ch: n.tau == 0 if it is the first iteration or if on the previous iteration no modification of the Hessian was necessary. In that case, if minA < 0, modify tau. If the Hessian was modified on the previous iteration, assume that a similar modification will be necessary at the current iteration and so reuse n.tau directly instead of starting from -minA + 0.001.

The value 0.001 is taken from Nocedal&Wright. I can make it a settable parameter, but I don't know how to name and document it succinctly and what the recommendation for its values should be.

vladimir-ch: Well, in fact, I would know how to name and document it, it is just not a terribly important parameter and the number of factorizations can already be controlled with ModifiedNewton.Increase.

newton.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+)
+
+// Newton implements a modified Newton's method for Hessian-based unconstrained
+// minimization. It applies regularization when the Hessian is not positive
+// definite, and uses line-searches so it can converge to a local minimum from
```
btracey: I don't think the "uses line searches" is needed here. It's described below, and isn't relevant to the important information (the rest of the paragraph)

newton.go
```diff
@@ -0,0 +1,143 @@
...
+// Newton iteratively forms a quadratic model to the objective function f and
+// tries to minimize this approximate model. It generates a sequence of
+// locations x_k by means of
+//  solve H_k d_k = -∇f_k for d_k,
+//  x_{k+1} = x_k + α_k d_k,
+// where H_k is the Hessian matrix of f at x_k and α_k is a step size found by
+// a line search.
+//
+// Away from a minimizer H_k may not be positive definite and d_k may not be a
+// descent direction. Newton implements a Hessian modification strategy that
```
btracey: that adds successively

newton.go
```diff
@@ -0,0 +1,143 @@
...
+// tries to add successively larger multiples of identity to H_k until it
+// becomes positive definite. Note that the repeated trial factorization of the
+// modified Hessian involved in this process can be computationally expensive.
+//
+// If the Hessian matrix cannot be formed explicitly or if the computational
+// cost of its factorization is prohibitive, BFGS or L-BFGS quasi-Newton method
+// can be used instead.
+type Newton struct {
+	// LinesearchMethod is a method used for selecting suitable steps along the
+	// descent direction d. Steps should satisfy the Wolfe, Goldstein or Armijo
```
btracey: should satisfy at least one of the

newton.go
```diff
@@ -0,0 +1,143 @@
...
+	// conditions. If LinesearchMethod == nil, an appropriate default is
+	// chosen.
+	LinesearchMethod LinesearchMethod
+	// Increase is a factor by which a scalar tau is successively increased so
```
btracey: Increase is the factor by which

newton.go
```diff
@@ -0,0 +1,143 @@
...
+func (n *Newton) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
+	return n.linesearch.Iterate(loc, xNext)
+}
+
+func (n *Newton) InitDirection(loc *Location, dir []float64) (stepSize float64) {
+	dim := len(loc.X)
+	n.chol = resizeTriDense(n.chol, dim)
+	n.hess = resizeSymDense(n.hess, dim)
+	n.tau = 0
+	n.computeNextDir(loc, dir)
```
btracey: Why does this need its own method and not just call NextDirection directly?

newton.go
```diff
@@ -0,0 +1,143 @@
...
+	return 1
+}
+
+func (n *Newton) NextDirection(loc *Location, dir []float64) (stepSize float64) {
+	n.computeNextDir(loc, dir)
+	return 1
+}
+
+func (n *Newton) computeNextDir(loc *Location, dir []float64) {
```
btracey: We should cite Nocedal and Wright here since that's where you got the algorithm from.

newton.go
btracey: Where is this in N&W? My copy suggests using the Frobenius norm of the matrix to set tau.

btracey: Oh! The comment is misleading. The comment in the block applies to the missing "else", not to the code right there. I think it's better to have a comment above the block. Something like
````
// If the smallest diagonal entry is positive, the Hessian may be positive definite, and so first attempt an un-modified cholesky. If the smallest entry is negative, use the final tau from the last iteration if regularization was needed, otherwise guess an appropriate value for tau
````

File: newton.go [hunk 14]
```diff
@@ -0,0 +1,143 @@
...
+	// that (H + tau*I) is positive definite. Larger values reduce the number
+	// of trial Hessian factorizations, but also reduce the second-order
+	// information in H.
+	// Increase must be greater than 1. If Increase is 0, it is defaulted to 5.
+	Increase float64
+
+	linesearch *Linesearch
+
+	hess *mat64.SymDense
+	chol *mat64.TriDense
```
btracey: Comment that this is storage for the cholesky.

newton.go
btracey: Sorry, I'm an idiot.

newton.go
```diff
@@ -0,0 +1,143 @@
...
+		// Try to apply the Cholesky decomposition.
+		pd := n.chol.Cholesky(n.hess, false)
+		if pd {
+			break
+		}
+		// Modified Hessian is not PD, so increase tau.
+		n.tau = math.Max(n.Increase*n.tau, 0.001)
+	}
+	d := mat64.NewDense(dim, 1, dir)
+	// Store the solution in d's backing array, dir.
```
btracey: Add a TODO that this should use mat64.Vector SolveCholesky when it exists (or, if you're feeling charitable, implement it :) ). This way we can use Dtrsv instead of Dtrsm

newton.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	for {
+		if n.tau != 0 {
+			// Add a multiple of identity to the Hessian.
+			for i := 0; i < dim; i++ {
+				n.hess.SetSym(i, i, loc.Hessian.At(i, i)+n.tau)
+			}
+		}
+		// Try to apply the Cholesky decomposition.
```
btracey: We should use the Upper cholesky here instead of the lower. I'm pretty sure it doesn't matter as far as the algorithm is concerned (since we're just using it to perform a solve next). However, SymDense is assumed to be UpperTriangular, and the eventual copy into the triangular is going to be faster into an upper triangular than a lower triangular.

types.go
```diff
@@ -257,3 +257,17 @@ func resize(x []float64, dim int) []float64 {
...
+func resizeSymDense(m *mat64.SymDense, dim int) *mat64.SymDense {
+	if m == nil || cap(m.RawSymmetric().Data) < dim*dim {
+		return mat64.NewSymDense(dim, nil)
+	}
+	return mat64.NewSymDense(dim, m.RawSymmetric().Data[:dim*dim])
+}
+
+func resizeTriDense(m *mat64.TriDense, dim int) *mat64.TriDense {
+	if m == nil || cap(m.RawTriangular().Data) < dim*dim {
+		return mat64.NewTriDense(dim, false, nil)
```
btracey: See above, I think this should be upper.

newton.go
vladimir-ch: I don't think so.

newton.go
vladimir-ch: Sorry for the confusion, I didn't give enough thought to where to put the comment.

Modified as you suggested.

newton.go
vladimir-ch: I'm not completely sure about this, but I think that for the theory a line search (together with some kind of Hessian modification) is necessary to get convergence from any starting point. The wording tried to reflect that.

I removed "uses line-searches so".

newton.go
vladimir-ch: Done

newton.go
vladimir-ch: Done

newton.go
vladimir-ch: Done

newton.go
vladimir-ch: Errrrrr ... you are right, of course. Fixed.

newton.go
vladimir-ch: Done

newton.go
vladimir-ch: Done

newton.go
vladimir-ch: Done

newton.go
vladimir-ch: Done

types.go
vladimir-ch: Done

newton.go
```diff
@@ -0,0 +1,145 @@
...
+	// the un-modified Hessian. If the smallest entry is negative, use the
+	// final tau from the last iteration if regularization was needed,
+	// otherwise guess an appropriate value for tau.
+	if minA > 0 {
+		n.tau = 0
+	} else if n.tau == 0 {
+		n.tau = -minA + 0.001
+	}
+
+	for {
```
vladimir-ch: I am not sure about this infinite loop. If `loc.Hessian` is reasonable, it will eventually break at line 122. But what if it contains some garbage? I can put a limit on the number of iterations here, but since NextDirection() does not return an error, what would be an appropriate response if the limit is reached? Just return the negative gradient? Comments, @btracey @kortschak ?

btracey: Yea, we need to do something. If the Hessian has an "inf" for example it will never conclude. Trying a few times and then doing the negative gradient seems reasonable. The other option is to return an error, but most users would probably rather find the optimum.

newton.go
```diff
@@ -0,0 +1,145 @@
...
+		if pd {
+			break
+		}
+		// Modified Hessian is not PD, so increase tau.
+		n.tau = math.Max(n.Increase*n.tau, 0.001)
+	}
+
+	d := mat64.NewDense(dim, 1, dir)
+	// Store the solution in d's backing array, dir.
+	// TODO(vladimir-ch): This should use mat64.Vector.SolveCholesky() when it exists.
```
btracey: Why is this a TODO? We have a Vector.SolveCholesky

:-)

newton.go
vladimir-ch: Ok, added `const maxNewtonModifications = 20`. The number is just a shot from the hip.

vladimir-ch:
Addressed all the issues and newly also added tests for Newton. Failing Travis build is again something with Travis, the tests all pass. PTAL, @btracey

newton.go
vladimir-ch: Oooh, you are so fast :-)

btracey:
LGTM

vladimir-ch:
I can't believe it, the Newton odyssey is over! And originally I thought it would be just a couple of commits. Thanks for bearing with me @btracey @kortschak .

Ta ta taaa (the one from The Croods)

PR #100, opened 2015-05-07
Update tests

vladimir-ch:
* Use `Method.Needs()` in tests instead of relying on what the objective function can evaluate
* Remove `TestLocalGradFree()`. Instead, update existing `TestLocal()` to use a mix of functions with or without `Grad()` method
* Use the same mix of functions in `TestNelderMead()`
* Take Hessian information into account in the tests

I'd like to review and merge this before #99 so that the tests for Newton can be added and merged as a part of that PR. When you have a chance, PTAL @btracey .


unconstrained_test.go
```diff
@@ -663,55 +726,71 @@ func testLocal(t *testing.T, tests []unconstrainedTest, method Method) {
...
 
-		// We are going to restart the solution using a fixed starting gradient
-		// and value, so evaluate them.
+		if method == nil {
+			// The tests below make sense only if the method used is known.
+			continue
+		}
+
+		if !method.Needs().Gradient && !method.Needs().Hessian {
+			// At present, gradient-free methods can correctly terminate only
```
btracey: This comment is not true. They can also converge with an error or if the maximum function evaluations is reached. The comment should state that "gradient-free tests can correctly terminate..."

btracey:
LGTM with comment fixed.

unconstrained_test.go
vladimir-ch: Fixed

PR #101, opened 2015-05-08
Add Hess() method to some test functions

vladimir-ch:
Another PR that I'd like to review and merge before #99 so that tests for Newton can go in as a part of that PR.

I chose simple and low-dimensional functions (except for one) and I used MINPACK-2 as a reference implementation, so I believe the implementation should be correct.

I also slightly modified the known minimum for `BrownAndDennis` function. The function value is higher on the level of floating-point noise, but the norm of the gradient is lower than before. I remember that inaccurate locations of minima were forcing us to use loose tolerances for the norm of the gradient in the past.

### DIFF
--- a/functions/functions.go
+++ b/functions/functions.go
@@ -595,11 +595,51 @@ func (BrownAndDennis) Grad(x, grad []float64) {
 	}
 }
 
+func (BrownAndDennis) Hess(x []float64, hess *mat64.SymDense) {
+	if len(x) != 4 {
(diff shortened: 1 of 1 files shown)


vladimir-ch:
PTAL, @btracey

btracey:
LGTM, assuming the travis fail isn't real

vladimir-ch:
It's not real. I've learned to ignore when travic-ci/pr fails and travis-ci/push passes, it is a common situation.

PR #105, opened 2015-05-18
Clean Printer and change Recorder interface

vladimir-ch:
* Make optimize silent by default
* Simplify Printer
* Change the printout format so that is is as fixed-width as possible for a majority of optimization runs.
* Remove FunctionInfo argument from Recorder.Init(). There is no reason why a Recorder should need information about the objective function for its initialization

PTAL @btracey

### DIFF
--- a/printer.go
+++ b/printer.go
@@ -25,7 +25,7 @@ var printerHeadings = [...]string{
 }
 
 const (
-	printerBaseTmpl = "%6v  %12v  %9v  %22v" // Base template for headings and values that are always printed.
+	printerBaseTmpl = "%9v  %16v  %9v  %22v" // Base template for headings and values that are always printed.
 	printerGradTmpl = "  %9v  %22v"          // Appended to base template when loc.Gradient != nil.
 	printerHessTmpl = "  %9v"                // Appended to base template when loc.Hessian != nil.
 )
--- a/printer.go
+++ b/printer.go
@@ -9,22 +9,35 @@ import (
 	"io"
 	"math"
 	"os"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/gonum/floats"
 )
 
+var printerHeadings = [...]string{
+	"Iter",
+	"Runtime",
+	"FuncEvals",
+	"Func",
+	"GradEvals",
+	"|Gradient|∞",
+	"HessEvals",
+}
+
+const (
+	printerBaseTmpl = "%6v  %12v  %9v  %22v" // Base template for headings and values that are always printed.
+	printerGradTmpl = "  %9v  %22v"          // Appended to base template when loc.Gradient != nil.
+	printerHessTmpl = "  %9v"                // Appended to base template when loc.Hessian != nil.
+)
+
 // Printer writes column-format output to the specified writer as the optimization
-// progresses. By default, it writes to Stdout.
+// progresses. By default, it writes to os.Stdout.
 type Printer struct {
 	Writer          io.Writer
 	HeadingInterval int
 	ValueInterval   time.Duration
 
 	lastHeading int
-	printGrad   bool
 	lastValue   time.Time
 }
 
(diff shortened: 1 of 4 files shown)


printer.go
```diff
@@ -35,113 +47,58 @@ func NewPrinter() *Printer {
...
-	}
-)
-
-func (p *Printer) Init(f *FunctionInfo) error {
-	p.printGrad = f.IsFunctionGradient || f.IsGradient
-
-	p.lastHeading = p.HeadingInterval + 1          // So the headings are printed the first time
-	p.lastValue = time.Now().Add(-p.ValueInterval) // So the values are printed the first time
+func (p *Printer) Init() error {
+	p.lastHeading = p.HeadingInterval
```
btracey: I would keep the comments here. Particularly the second line is very weird if you don't understand the purpose.

printer.go
```diff
@@ -9,13 +9,26 @@ import (
...
 )
 
+var printerHeadings = [...]string{
+	"Iter",
+	"Runtime",
+	"FuncEvals",
+	"GradEvals",
+	"HessEvals",
+	"Func",
+	"|Gradient|",
```
btracey: Could we add an infinity symbol after the norm?

btracey: Unless you think the non-ascii will break stuff.

printer.go
```diff
@@ -9,13 +9,26 @@ import (
...
+}
+
+const (
```
btracey: These need comments.

printer.go
```diff
@@ -35,113 +47,58 @@ func NewPrinter() *Printer {
...
 
-	var maxLengths [nPrinterOut]int
-
-	for i := 0; i < nPrint; i++ {
-		v := len(printerHeadings[i])
-		v2 := len(valueStrings[i])
-		if v > v2 {
-			maxLengths[i] = v
-			continue
+	if loc.Gradient == nil {
```
btracey: This style isn't scalable. When constraints come, we'll also (maybe) have constraint violation and constraint gradient. It's preferable to start the string with the set of standard outputs (runtime, iter, etc.), and then append.

printer.go
vladimir-ch: Done

printer.go
vladimir-ch: Done.

I don't think non-ascii is a problem, printing microseconds in default format already uses μ anyway.

printer.go
vladimir-ch: Done

printer.go
vladimir-ch: Changed

vladimir-ch:
PTAL

printer.go
```diff
@@ -9,22 +9,35 @@ import (
...
+	"Runtime",
+	"FuncEvals",
+	"Func",
+	"GradEvals",
+	"|Gradient|∞",
+	"HessEvals",
+}
+
+const (
+	printerBaseTmpl = "%6v  %12v  %9v  %22v" // Base template for headings and values that are always printed.
```
btracey: I don't understand the choices of length here.

btracey:
LGTM except for the question about lengths.

printer.go
vladimir-ch: The overall intention was to have a fixed-width output for the majority of runs.

Width of 6 for the iteration may be slightly excessive, I can make it 5 or even 4 if you want.

Width of 12 is just an empirical observation of the maximum string length that the default format for Time gives.

Width of 9 for the number of evaluations is simply the width of the headings `FuncEvals`, ...  It is excessive, so if you can think of a shorter, descriptive heading, I will replace it.

Width of 22 for floats is again an empirical observation: 16 digits + sign, dot, e, sign of the exponent and 2 digits of the exponent. We could use a real fixed-width format like %12.5e but %v has proved useful for debugging in the past.

vladimir-ch: I reduced the default width for iterations to 5. I also had a look at time.Duration.String() and I increased the default width for Runtime to 16 which is enough for a fixed-width output up to 1 hour. PTAL, @btracey

vladimir-ch:
A very gentle ping: Does this still have your LGTM with the modified default widths, @btracey ?

btracey:
I'd go longer on iter (my maximum run so far was 2e6 iterations). Make it 9 digits long. That way all integers are 9 and all floats are 22. Then LGTM

PR #106, opened 2015-05-18
Update MinimalSurface to implement Gradient interface

vladimir-ch:
When FunctionGradient interface was removed, MinimalSurface function was not updated. This PR fixes that. PTAL @btracey


btracey:
LGTM

PR #107, opened 2015-05-28
Fix BFGS to have vector arguments in Inner.

btracey:
mat64 changed so that Inner takes a *mat64.Vector instead of a []float64. Update BFGS to this new API


btracey:
Merging because fixes the build.

PTAL @vladimir-ch

kortschak:
LGTM

Do we have any other cases of this - stat?

btracey:
I don't think it's anywhere else. It's not used in stat.

PR #109, opened 2015-06-08
Switch from interfaces to Problem struct

hypoactiv:
My interpretation of Issue 102. Comments and criticism welcome.

### DIFF
--- a/backtracking.go
+++ b/backtracking.go
@@ -32,7 +32,7 @@ type Backtracking struct {
 	initG    float64
 }
 
-func (b *Backtracking) Init(loc LinesearchLocation, step float64, _ *FunctionInfo) EvaluationType {
+func (b *Backtracking) Init(loc LinesearchLocation, step float64, _ *ProblemInfo) EvaluationType {
 	if step <= 0 {
 		panic("backtracking: bad step size")
 	}
--- a/cg.go
+++ b/cg.go
@@ -94,7 +94,7 @@ type CG struct {
 	gradPrevNorm float64
 }
 
-func (cg *CG) Init(loc *Location, probInfo *ProblemInfo, xNext []float64) (EvaluationType, IterationType, error) {
+func (cg *CG) Init(loc *Location, p *ProblemInfo, xNext []float64) (EvaluationType, IterationType, error) {
 	if cg.IterationRestartFactor < 0 {
 		panic("cg: IterationRestartFactor is negative")
 	}
@@ -125,7 +125,7 @@ func (cg *CG) Init(loc *Location, probInfo *ProblemInfo, xNext []float64) (Evalu
 	cg.linesearch.Method = cg.LinesearchMethod
 	cg.linesearch.NextDirectioner = cg
 
-	return cg.linesearch.Init(loc, probInfo, xNext)
+	return cg.linesearch.Init(loc, p, xNext)
 }
 
 func (cg *CG) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
--- a/cg.go
+++ b/cg.go
@@ -94,7 +94,7 @@ type CG struct {
 	gradPrevNorm float64
 }
 
-func (cg *CG) Init(loc *Location, funcInfo *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
+func (cg *CG) Init(loc *Location, probInfo *ProblemInfo, xNext []float64) (EvaluationType, IterationType, error) {
 	if cg.IterationRestartFactor < 0 {
 		panic("cg: IterationRestartFactor is negative")
 	}
@@ -125,7 +125,7 @@ func (cg *CG) Init(loc *Location, funcInfo *FunctionInfo, xNext []float64) (Eval
 	cg.linesearch.Method = cg.LinesearchMethod
 	cg.linesearch.NextDirectioner = cg
 
-	return cg.linesearch.Init(loc, funcInfo, xNext)
+	return cg.linesearch.Init(loc, probInfo, xNext)
 }
 
 func (cg *CG) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
--- a/interfaces.go
+++ b/interfaces.go
@@ -4,26 +4,6 @@
 
 package optimize
 
-import "github.com/gonum/matrix/mat64"
-
-// Function evaluates the objective function at the given location. F
-// must not modify x.
-type Function interface {
-	Func(x []float64) (obj float64)
-}
-
-// Gradient evaluates the gradient at x and stores the result in-place in grad.
-// Grad must not modify x.
-type Gradient interface {
-	Grad(x, grad []float64)
-}
-
-// Hessian evaluates the Hessian at x and stores the result in-place in hess.
-// Hess must not modify x.
-type Hessian interface {
-	Hess(x []float64, hess *mat64.SymDense)
-}
-
 // LinesearchMethod is a type that can perform a line search. Typically, these
 // methods will not be called by the user directly, as they will be called by
 // a Linesearch struct.
--- a/interfaces.go
+++ b/interfaces.go
@@ -11,7 +11,7 @@ type LinesearchMethod interface {
 	// Init initializes the linesearch method. LinesearchLocation contains the
 	// function information at step == 0, and step contains the first step length
 	// as specified by the NextDirectioner.
-	Init(loc LinesearchLocation, step float64, f *FunctionInfo) EvaluationType
+	Init(loc LinesearchLocation, step float64, p *ProblemInfo) EvaluationType
 
 	// Finished takes in the function result at the most recent linesearch location,
 	// and returns true if the line search has been concluded.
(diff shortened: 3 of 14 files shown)


hypoactiv:
The calling style for the modified code is
```
	p := optimize.Problem{
		Func: func(x []float64) float64 {
			return math.Pow(x[0]*x[0]-2, 2)
		},
		Grad: func(x []float64, grad []float64) {
			grad[0] = 4 * x[0] * (x[0]*x[0] - 2)
		},
		Hess: func(x []float64, hess *mat64.SymDense) {
			hess.SetSym(0, 0, 4*(x[0]*x[0]-2)+8*x[0]*x[0])
		},
	}

	res, err := optimize.Local(p, []float64{1}, nil, &optimize.Newton{})
```

backtracking.go
```diff
@@ -32,7 +32,7 @@ type Backtracking struct {
 	initG    float64
 }
 
-func (b *Backtracking) Init(loc LinesearchLocation, step float64, _ *FunctionInfo) EvaluationType {
+func (b *Backtracking) Init(loc LinesearchLocation, step float64, _ *Problem) EvaluationType {
```
btracey: I know FunctionInfo seems redundant, but it's not. A big invariant in this package is that optimization methods cannot call the functions directly. Giving them access through Problem makes it very tempting to cheat, which breaks all measurement (of things like NumEvals)

types.go
```diff
@@ -236,6 +206,14 @@ type Settings struct {
 	Recorder Recorder
 }
 
+// TODO Comment
```
btracey: I think we should decide on the full Problem struct now (with constraints), as that way we can hope we don't need more incompatible changes. The two outstanding issues in my mind are how to communicate the number of nonlinear constraints and if Hess et. al. should take matrix interface types instead of concrete types. Thoughts @vladimir-ch ?

vladimir-ch: Do we really have to decide the full Problem struct now? There is no code that needs constraints now and apparently it is unclear how the nonlinear constraints should be represented.

How about having two functions, one for equality and one for inequality constraints, that return a slice of functions? Similarly for their gradient, but then it gets a bit more complicated. I am not sure, but having an access to individual constraints functions could be useful. I don't understand your remark about calling Constraints with nil arguments the first time, but whatever it means it doesn't sound like something fancy.

I would say that Hess et al. should take matrix interfaces, it would leave the door open for any future implementation of sparse problems when there are sparse symmetric matrices, for example. Unless there are some drawbacks that I'm not aware of.

local.go
```diff
@@ -49,13 +49,13 @@ import (
 // function evaluations. If you would like to put limits on this, for example
 // maximum runtime or maximum function evaluations, please modify the Settings
 // input struct.
-func Local(f Function, initX []float64, settings *Settings, method Method) (*Result, error) {
+func Local(p Problem, initX []float64, settings *Settings, method Method) (*Result, error) {
 	if len(initX) == 0 {
 		panic("optimize: initial X has zero length")
 	}
```
vladimir-ch: We should panic here if p.Func is nil

types.go
```diff
@@ -108,20 +108,16 @@ type FunctionInfo struct {
 
 // functionInfo contains information about which interfaces the objective
 // function F implements and the actual methods of F that have been
-// successfully type switched.
+// successfully type switched. TODO(jds) change this comment
 type functionInfo struct {
```
vladimir-ch: I think this struct existed so that we didn't have to typeswitch on Function all the time. Now that we are explicit, this struct is probably unnecessary. And we should check if FunctionInfo (possibly renamed to ProblemInfo) is absolutely necessary or if it could be removed as well. @btracey explained some time ago why the optimizers could want it, but I forgot the reason and the package has changed since then.

vladimir-ch: In any case, none of the current optimizers actually need FunctionInfo that we pass to it.

btracey: I think I had two imagined use cases. The first was that the method could test that it could handle the kind of optimization problem. We implemented that with the Needs method instead.

Another possible use would be in Global, having a "random restart Local method". If it wanted to choose a default local method, it would need to know what functions were non-nil.

It seems like there would be optimizers that would change behavior dependent on what kind of problem it is.

vladimir-ch: Sure, I don't object to having FunctionInfo, I just needed a gentle reminder why we have it. So we keep it, but I would rename it to ProblemInfo and the fields from IsGradient to HasGradient etc. "Is" came from Function _being_ Gradient, etc interface.

hypoactiv:
If this isn't too far from what was in mind with 102, I'd like to start working on converting the test functions to the new Problem struct. My first idea of how to do this was to define a Problem struct for each test function in functions.go. This creates an import loop though, since functions.go needs to import optimize to get the Problem struct and unconstrained_test.go needs to import the test functions package.

My next idea was to create a struct in functions.go (TestProblem, or something similar) that contains the objective function, gradient, etc. for each test problem. This could then be converted into a Problem in unconstrained_test.go before running the test. I'm not very enthusiastic about this idea though, since TestProblem contains essentially the same information as Problem.

btracey:
I don't think you need to go to a huge change. Just create a Problem using, for example, ExtendedRosenbrock{}.Func. In general, that sort of style may be best for many users of the code, so it may be a good idea to provide such example in the test suite.

vladimir-ch:
Thanks for working on this, @hypoactiv .

What is the status of this PR, is it ready for a review? Should I have look?

hypoactiv:
I think it's ready for you to take a look, @vladimir-ch. I'd appreciate your input. Thanks!

cg.go
```diff
@@ -94,7 +94,7 @@ type CG struct {
 	gradPrevNorm float64
 }
 
-func (cg *CG) Init(loc *Location, funcInfo *FunctionInfo, xNext []float64) (EvaluationType, IterationType, error) {
+func (cg *CG) Init(loc *Location, probInfo *ProblemInfo, xNext []float64) (EvaluationType, IterationType, error) {
```
vladimir-ch: Let's use `p` for `ProblemInfo` arguments consistently.

interfaces.go
```diff
@@ -4,26 +4,6 @@
 
 package optimize
 
-import "github.com/gonum/matrix/mat64"
-
-// Function evaluates the objective function at the given location. F
```
vladimir-ch: These interface comments should be moved to `Problem`.

types.go
```diff
@@ -97,52 +97,44 @@ type Stats struct {
...
-	IsGradient                bool
-	IsHessian                 bool
-	IsFunctionGradient        bool
-	IsFunctionGradientHessian bool
-	IsStatuser                bool
+// ProblemInfo is data to give to the optimizer about the objective function.
+type ProblemInfo struct {
+	HasGradient                bool
+	HasHessian                 bool
+	HasFunctionGradient        bool
```
vladimir-ch: This field and the next one should be removed. We don't have these combined evaluation methods anymore.

unconstrained_test.go
```diff
@@ -16,8 +15,10 @@ import (
 )
 
 type unconstrainedTest struct {
-	// f is the function that is being minimized.
-	f Function
+	// name is the name of the test function.
```
vladimir-ch: I would say that it is the name of the test, not only the test function.

unconstrained_test.go
```diff
@@ -776,11 +1189,11 @@ func testLocal(t *testing.T, tests []unconstrainedTest, method Method) {
 			continue
 		}
 
-		f := newFunctionInfo(test.f)
+		p := newProblemInfo(test.p)
```
vladimir-ch: I think this is not necessary anymore, it existed to avoid the type assertions. You can use `test.p` directly below.

types.go
```diff
@@ -97,52 +97,44 @@ type Stats struct {
...
-type functionInfo struct {
-	FunctionInfo
-
-	function Function
-	gradient Gradient
-	hessian  Hessian
-	statuser Statuser
+// problemInfo contains information about which methods the optimization
+// problem provides as well as the optimization problem itself.
+type problemInfo struct {
```
vladimir-ch: I think we should remove this type. It existed to avoid repeated type assertions on `Function`, now we can easily test fields of `Problem` for nil. Internal functions that take `problemInfo` can take `Problem` directly. `satisfies` can be an unexported method on `Problem` until we make it exported or find a better name and place for it. The other PR in queue removes passing of `ProblemInfo` to the `Recorder`, so after it is merged `problemInfo` won't be needed in `Local()`.

vladimir-ch: So the other PR has been merged. You should update your branch and the code accordingly.

vladimir-ch:
Ping

hypoactiv:
Sorry for disappearing like that -- school and then a reinstall of Linux had me distracted.

I implemented your suggestions @vladimir-ch. When I was removing `problemInfo` I kept `newProblemInfo` and converted it to construct the `ProblemInfo` we pass to the optimization method. I also switched `satisfies` to be on `Problem`, as suggested.

types.go
```diff
@@ -97,55 +97,23 @@ type Stats struct {
...
-	IsGradient                bool
-	IsHessian                 bool
-	IsFunctionGradient        bool
-	IsFunctionGradientHessian bool
-	IsStatuser                bool
+// ProblemInfo is data to give to the optimizer about the objective function.
+type ProblemInfo struct {
+	HasGradient bool
+	HasHessian  bool
+	HasStatus   bool
```
vladimir-ch: I think that we can also remove `HasStatus`, it is a piece of information that is useless to `Method`.

vladimir-ch:
> Sorry for disappearing like that -- school and then a reinstall of Linux had me distracted.

No problem at all. Thanks for your contribution.

After you have addressed the last two comments (hasXXX variables and HasStatus field), this PR has my LGTM. Then I will merge it into master.

btracey:
Add your name to authors

Issue #110, opened 2015-07-17
Method does not need ProblemInfo


vladimir-ch:
I have been giving it more thought and I am still not convinced that passing `ProblemInfo` to `Method.Init()` makes sense. Methods already announce what they need via `Needs()` and we allocate fields of `Location` exactly according to this information. Evaluation types issued by Method should/have to match this information. So ProblemInfo will convey what Needs() returned plus perhaps something more but Method cannot use that information anyway, it cannot issue evaluation types beyond Needs(). Comments, @btracey ?

btracey:
My thought on having it was that hypothetically Methods could have different behavior depending on the ProblemInfo. I'm not sure I have a good example though. The one I had in mind was random restart global optimization. In this case though, the Restarter struct would have something like
````
type Restarter struct{
    Local Method   // The local optimization method for each restart
}
````
Here though, Restarter.Needs() would just return Local.Needs(), and so it doesn't need ProblemInfo.

PR #111, opened 2015-07-27
Remove unnecessary ProblemInfo

vladimir-ch:
Discussion in #110 implies that `Method` cannot make any use of `ProblemInfo` which therefore becomes unnecessary.

PTAL, @btracey

### DIFF
--- a/backtracking.go
+++ b/backtracking.go
@@ -32,7 +32,7 @@ type Backtracking struct {
 	initG    float64
 }
 
-func (b *Backtracking) Init(loc LinesearchLocation, step float64, _ *ProblemInfo) EvaluationType {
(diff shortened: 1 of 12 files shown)


btracey:
LGTM. I'm not convinced we don't need it, but we can add it back in if we do.

vladimir-ch:
I understand your hesitation, but I really think that ProblemInfo does nor provide any information at all, it just reflects what Needs() returns. And if Problem implements more than what Needs() returns, Method is not able to call it because the corresponding fields of Location are not allocated. Perhaps in the future we will need a more flexible mechanism than this, then we can revisit it, but for now I think it is cleaner to remove ProblemInfo.

PR #112, opened 2015-07-28
Clean and improve Linesearch code

vladimir-ch:
This is something I have wanted to do for some time now:

* `Linesearch.Init()` and `Linesearch.initNextLinesearch()` were both doing almost exactly the same thing. Now the former calls the latter.
* It is very unlikely that `Linesearch` could be used for a non-gradient-based optimization. Init() now panics, if `loc.Gradient == nil`, which makes the code simpler.
* Both `Linesearch.Iterate()` and `Linesearch.initNextLinesearch()` were slightly cleaned and improved, for example `evalType` and `iterType` are now both consistently updated.
* Docs for `Linesearch` were improved to make clear that it is just a helper struct, not a method in itself.
* I renamed `Linesearch` to `LinesearchHelper` and `LinesearchMethod` to `Linesearch`. As the commit message says:

> Linesearch is a too nice name to be used for only a helper struct.
LinesearchHelper better transmits its purpose and leaves the name
Linesearch free for what used to be LinesearchMethod. LinesearchMethod
was a bit confusing because Method is the interface for any optimization
method and so the name suggested a relation that was not there.

* I also renamed the related fields in GradientDescent, CG, BFGS, LBFGS and Newton to match the interface names. The previous situation was constantly distracting me.
* Removed `LinesearchLocation`

The diffs are dense, but there is nothing special going on except for many renames (mostly). The commits are better viewed separately.

@btracey , PTAL

~~There is one more thing that I would like to add to this PR to take it all with one strike. I don't like `LinesearchLocation` very much. The name is too long and the type's sole purpose is to pass two numbers to `Linesearch` (former `LinesearchMethod`). That can be done by passing the numbers directly as arguments. So if you agree with eliminating `LinesearchLocation` (the name is not very accurate anyway) @btracey , I would like to add this change too.~~

### DIFF
--- a/backtracking.go
+++ b/backtracking.go
@@ -5,13 +5,13 @@
 package optimize
 
 const (
-	defaultBacktrackingDecrease = 0.5
-	defaultBacktrackingFunConst = 1e-4
-	minimumBacktrackingStepSize = 1e-20
+	defaultBacktrackingDecrease  = 0.5
+	defaultBacktrackingFuncConst = 1e-4
+	minimumBacktrackingStepSize  = 1e-20
 )
 
 // Backtracking is a Linesearcher that uses a backtracking to find a point that
-// satisfies the Armijo condition with the given function constant FunConst. If
+// satisfies the Armijo condition with the given function constant FuncConst. If
 // the Armijo condition has not been met, the step size is decreased by a
 // factor of Decrease.
 //
@@ -21,12 +21,12 @@ const (
 // not appropriate for optimizers that require the Wolfe conditions to be met,
 // such as BFGS.
 //
-// Both FunConst and Decrease must be between zero and one, and Backtracking will
-// panic otherwise. If either FunConst or Decrease are zero, it will be set to a
+// Both FuncConst and Decrease must be between zero and one, and Backtracking will
+// panic otherwise. If either FuncConst or Decrease are zero, it will be set to a
 // reasonable default.
 type Backtracking struct {
-	FunConst float64 // Necessary function descrease for Armijo condition.
-	Decrease float64 // Step size multiplier at each iteration (stepSize *= Decrease).
+	FuncConst float64 // Necessary function descrease for Armijo condition.
+	Decrease  float64 // Step size multiplier at each iteration (stepSize *= Decrease).
 
 	stepSize float64
 	initF    float64
@@ -44,14 +44,14 @@ func (b *Backtracking) Init(f, g float64, step float64) EvaluationType {
 	if b.Decrease == 0 {
 		b.Decrease = defaultBacktrackingDecrease
 	}
-	if b.FunConst == 0 {
-		b.FunConst = defaultBacktrackingFunConst
+	if b.FuncConst == 0 {
+		b.FuncConst = defaultBacktrackingFuncConst
 	}
 	if b.Decrease <= 0 || b.Decrease >= 1 {
 		panic("backtracking: Decrease must be between 0 and 1")
 	}
-	if b.FunConst <= 0 || b.FunConst >= 1 {
-		panic("backtracking: FunConst must be between 0 and 1")
+	if b.FuncConst <= 0 || b.FuncConst >= 1 {
+		panic("backtracking: FuncConst must be between 0 and 1")
 	}
 
 	b.stepSize = step
@@ -61,7 +61,7 @@ func (b *Backtracking) Init(f, g float64, step float64) EvaluationType {
 }
 
 func (b *Backtracking) Finished(f, _ float64) bool {
-	return ArmijoConditionMet(f, b.initF, b.initG, b.stepSize, b.FunConst)
+	return ArmijoConditionMet(f, b.initF, b.initG, b.stepSize, b.FuncConst)
 }
 
 func (b *Backtracking) Iterate(_, _ float64) (float64, EvaluationType, error) {
--- a/backtracking.go
+++ b/backtracking.go
@@ -10,15 +10,16 @@ const (
 	minimumBacktrackingStepSize = 1e-20
 )
 
-// Backtracking is a type that implements LinesearchMethod using a backtracking
-// line search. A backtracking line search checks that the Armijo condition has
-// been met with the given function constant. If the Armijo condition has not
-// been met, the step size is decreased by a factor of Decrease.
+// Backtracking is a Linesearcher that uses a backtracking to find a point that
+// satisfies the Armijo condition with the given function constant FunConst. If
+// the Armijo condition has not been met, the step size is decreased by a
+// factor of Decrease.
 //
-// The Armijo conditions only require the gradient at the initial condition
-// (not successive step locations), and so Backtracking may be a good linesearch
-// method for functions with expensive gradients. Backtracking is not appropriate
-// for optimizers that require the Wolfe conditions to be met, such as BFGS.
+// The Armijo condition only requires the gradient at the beginning of each
+// major iteration (not at successive step locations), and so Backtracking may
+// be a good linesearch for functions with expensive gradients. Backtracking is
+// not appropriate for optimizers that require the Wolfe conditions to be met,
+// such as BFGS.
 //
 // Both FunConst and Decrease must be between zero and one, and Backtracking will
 // panic otherwise. If either FunConst or Decrease are zero, it will be set to a
--- a/backtracking.go
+++ b/backtracking.go
@@ -10,7 +10,7 @@ const (
 	minimumBacktrackingStepSize  = 1e-20
 )
 
-// Backtracking is a Linesearcher that uses a backtracking to find a point that
+// Backtracking is a Linesearcher that uses backtracking to find a point that
 // satisfies the Armijo condition with the given function constant FuncConst. If
 // the Armijo condition has not been met, the step size is decreased by a
 // factor of Decrease.
--- a/backtracking.go
+++ b/backtracking.go
@@ -33,11 +33,11 @@ type Backtracking struct {
 }
 
 func (b *Backtracking) Init(f, g float64, step float64) EvaluationType {
-	if g >= 0 {
-		panic("backtracking: initial derivative is non-negative")
-	}
 	if step <= 0 {
 		panic("backtracking: bad step size")
+	}
+	if g >= 0 {
+		panic("backtracking: initial derivative is non-negative")
 	}
 
 	if b.Decrease == 0 {
--- a/backtracking.go
+++ b/backtracking.go
@@ -32,12 +32,12 @@ type Backtracking struct {
 	initG    float64
 }
 
-func (b *Backtracking) Init(loc LinesearchLocation, step float64) EvaluationType {
+func (b *Backtracking) Init(f, g float64, step float64) EvaluationType {
+	if g >= 0 {
+		panic("backtracking: initial derivative is non-negative")
+	}
 	if step <= 0 {
 		panic("backtracking: bad step size")
-	}
-	if loc.Derivative >= 0 {
-		panic("Backtracking: init G non-negative")
 	}
 
 	if b.Decrease == 0 {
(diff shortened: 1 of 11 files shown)


btracey:
I agree with removing LinesearchLocation. The original thought was that there could be one linesearch (constrained, unconstrained, derivative, derivative free). I agree that this original thought overcomplicates matters. The basic linesearch can be gradient based, and if we need other types we can add them.

vladimir-ch:
Ok, I removed `LinesearchLocation` and updated the interface docs. This is ready for a review, PTAL, @btracey .

backtracking.go
```diff
@@ -32,13 +32,13 @@ type Backtracking struct {
 	initG    float64
 }
 
-func (b *Backtracking) Init(loc LinesearchLocation, step float64) EvaluationType {
+func (b *Backtracking) Init(f, g float64, step float64) EvaluationType {
+	if g >= 0 {
+		panic("backtracking: initial derivative is non-negative")
+	}
```
btracey: Could you move this after the step check? I can see the case where someone got both signs wrong, so the step is negative and the gradient is positive. Seeing the step check first is more useful.

interfaces.go
```diff
@@ -4,28 +4,54 @@
...
+}
+
+// Linesearch is a type that can perform a line search. It tries to find an
+// (approximate) minimum of the objective function along the search direction
+// dir_k starting at the most recent location x_k, i.e., it tries to minimize a
+// function
+//  φ(step) := f(x_k + step * dir_k), step > 0.
+// Typically, these methods will not be called by the user directly, as they
+// will be called by the LinesearchHelper struct.
+type Linesearch interface {
```
btracey: I think this would be nicer called "Linesearcher", meaning something that can perform a linesearch. The interface doesn't fall into the strict go rules, but I think the code reads a little more clearly with the "er".

linesearch.go
```diff
@@ -10,123 +10,125 @@ import (
 	"github.com/gonum/floats"
 )
 
-// Linesearch is a linesearch-based optimization method.
-// It consists of a NextDirectioner, which specifies the next linesearch method,
-// and a LinesearchMethod which performs the linesearch in the direction specified
-// by the NextDirectioner.
-type Linesearch struct {
+// LinesearchHelper encapsulates the common functionality of gradient-based
```
btracey: How about "LinesearcherGrad" instead?

linesearch.go
```diff
@@ -10,123 +10,125 @@ import (
...
+// line-search optimization methods and serves as a helper struct for their
```
btracey: remove optimization

linesearch.go
btracey: Nope. This name is better.

btracey:
Sorry for the scattered comments. This is the problem without being able to submit all at once.

I do not think it is sufficiently clear that LinesearchHelper is really an optimization method. This was a problem before this PR, but I think now is the time to fix it. BFGS, for example, is generally described as "BFGS sets a direction, then it performs a linesearch, and then it tests for convergence". The control is inverted in gonum/optimize. We take the view that we are optimizing with successive linesearches, and BFGS et al. are providing the next direction in which to linesearch.  We should choose a set of names that make this behavior clear.

I originally chose "LinesearchMethod" in an attempt to highlight that fact, but that name was chosen for the wrong struct. Instead, I think LinesearchMethod should become Linesearcher. It is a type that can perform a linesearch. Then, the current Linesearch (your LinesearchHelper) becomes LinesearchMethod. The documentation then also changes. It is not a helper type for implementing gradient-based line searches, it is an optimization method that optimizes via sequential gradient-based line searches. 

Thoughts?

kortschak:
>I originally chose "LinesearchMethod" in an attempt to highlight that fact, but that name was chosen for the wrong struct. Instead, I think LinesearchMethod should become Linesearcher. It is a type that can perform a linesearch. Then, the current Linesearch (your LinesearchHelper) becomes LinesearchMethod. The documentation then also changes. It is not a helper type for implementing gradient-based line searches, it is an optimization method that optimizes via sequential gradient-based line searches.

I was pondering the names a day ago, thinking that we could do better, but not knowing enough about the domain to really make a sensible suggestion. I think the suggestion of Linsearcher and LinesearchMethod works - though speaking from a position of profound ignorance.

vladimir-ch:
I wanted to change the names because they were inaccurate and utterly confusing even for me. Currently, we have `Method` interface. Then we have `LinesearchMethod` struct which does not implement that interface (because it is not an optimization method). Then we have `Linesearch` which is an abstract optimization method but still does not implement `Method` interface. I did not check but quite possibly in the docs "linesearch" is also used interchangably for line-search based optimization method and for the line-searcher part of those methods. So I wanted to find better type names.

`Linesearcher` is a great name for what was `LinesearchMethod`, it rhymes with `NextDirectioner`, so I will make that change.

`LinesearchMethod` is also nice and descriptive, only that the Method part could suggest that it implements `Method` which it does not, it does not have `Needs()`. The way we use it in the package is really just as a helper struct, none of the optimizers expose it. Sure, outside implementers do not have to use it the same way. All in all, if you think that not satisfying `Method` interface does not matter, I will change `LinesearchHelper` to `LinesearchMethod`.

backtracking.go
vladimir-ch: Done.

interfaces.go
vladimir-ch: And it fits nicely with NextDirectioner and StepSizer. Done.

linesearch.go
vladimir-ch: Done.

vladimir-ch:
Ok, so in the end I changed `LinesearchHelper` to `LinesearchMethod`. It is a nice name even though it does not implement `Method`.

I will merge related commits before merging.

PTAL, @btracey @kortschak

backtracking.go
```diff
@@ -5,65 +5,66 @@
...
+	defaultBacktrackingDecrease  = 0.5
+	defaultBacktrackingFuncConst = 1e-4
+	minimumBacktrackingStepSize  = 1e-20
 )
 
-// Backtracking is a type that implements LinesearchMethod using a backtracking
-// line search. A backtracking line search checks that the Armijo condition has
-// been met with the given function constant. If the Armijo condition has not
-// been met, the step size is decreased by a factor of Decrease.
+// Backtracking is a Linesearcher that uses a backtracking to find a point that
```
btracey: that uses backtracking

vladimir-ch: Done.

interfaces.go
```diff
@@ -4,28 +4,54 @@
...
+	// by Init() and Iterate() during the optimization process.
+	Needs() struct {
+		Gradient bool
+		Hessian  bool
+	}
+}
+
+// Linesearcher is a type that can perform a line search. It tries to find an
+// (approximate) minimum of the objective function along the search direction
+// dir_k starting at the most recent location x_k, i.e., it tries to minimize a
```
btracey: to minimize the function

interfaces.go
```diff
@@ -4,28 +4,54 @@
...
+// function
+//  φ(step) := f(x_k + step * dir_k), step > 0.
+// Typically, these methods will not be called by the user directly, as they
```
btracey: How about "Typically, a Linesearcher will be used in conjuction with LinesearchMethod for performing gradient-based optimization through sequential line searches."

interfaces.go
```diff
@@ -4,28 +4,54 @@
...
+// will be called by the LinesearchMethod struct.
+type Linesearcher interface {
+	// Init initializes the linesearch method. Value and derivative contain
+	// φ(0) and φ'(0), respectively, and step contains the first trial step
+	// length as returned by the NextDirectioner.InitDirection(). It returns
```
btracey: I don't like the implication that it has to come from NextDirectioner.InitDirection (maybe someone just wants to do a linesearch). Just say "contains the first trial step length".

interfaces.go
```diff
@@ -4,28 +4,54 @@
...
+	// and returns the next step size and the type of evaluation to be
+	// performed at x_k + step * dir_k.
+	Iterate(value, derivative float64) (step float64, e EvaluationType, err error)
 }
 
-// NextDirectioner implements a strategy for computing a new line search direction
-// at each major iteration. Typically, these methods will not be called by the user directly,
-// as they will be called by a Linesearch struct.
+// NextDirectioner implements a strategy for computing a new line search
+// direction at each major iteration. Typically, these methods will not be
```
btracey: Same comment as with Linesearcher

linesearch.go
```diff
@@ -10,132 +10,135 @@ import (
 	"github.com/gonum/floats"
 )
 
-// Linesearch is a linesearch-based optimization method.
-// It consists of a NextDirectioner, which specifies the next linesearch method,
-// and a LinesearchMethod which performs the linesearch in the direction specified
-// by the NextDirectioner.
-type Linesearch struct {
+// LinesearchMethod encapsulates the common functionality of gradient-based
```
btracey: I liked your description in the comments
Here would be my suggested comment.
"LinesearchMethod represents an abstract optimization method in which a function is optimized through successive line search optimizations. It consists of a NextDirectioner, which specifies the search direction at each iteration, and a Linesearcher which performs a linesearch along the search direction."

linesearch.go
```diff
@@ -10,132 +10,135 @@ import (
...
-	floats.AddScaledTo(xNext, ls.initX, stepSize, ls.dir)
-	ls.lastEvalType = evalType
-	ls.iterType = MinorIteration
-	return evalType, ls.iterType, nil
+	return ls.initNextLinesearch(loc, xNext)
 }
 
-func (ls *Linesearch) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
+func (ls *LinesearchMethod) Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
+	if ls.iterType == NoIteration {
```
btracey: I'm worried about the interaction between this and the case where no progress can be made. It seems that it would be hard to handle the error if the client has to call Init() again.

interfaces.go
vladimir-ch: Done.

interfaces.go
vladimir-ch: Nice. Done.

interfaces.go
vladimir-ch: Done.

interfaces.go
vladimir-ch: Done.

linesearch.go
vladimir-ch: Done.

linesearch.go
vladimir-ch: Without calling Init() the only thing the client can do to recover from no progress is something with the Linesearcher. With this check it is not possible, so I will remove it, there is not much harm in doing that.

vladimir-ch:
PTAL @btracey

interfaces.go
```diff
@@ -4,28 +4,54 @@
...
+		Gradient bool
+		Hessian  bool
+	}
+}
+
+// Linesearcher is a type that can perform a line search. It tries to find an
+// (approximate) minimum of the objective function along the search direction
+// dir_k starting at the most recent location x_k, i.e., it tries to minimize
+// the function
+//  φ(step) := f(x_k + step * dir_k), step > 0.
```
btracey: where instead of comma

linesearch.go
```diff
@@ -10,132 +10,130 @@ import (
...
 
-// Linesearch is a linesearch-based optimization method.
-// It consists of a NextDirectioner, which specifies the next linesearch method,
-// and a LinesearchMethod which performs the linesearch in the direction specified
-// by the NextDirectioner.
-type Linesearch struct {
+// LinesearchMethod represents an abstract optimization method in which
+// a function is optimized through successive line search optimizations.
+// It consists of a NextDirectioner, which specifies the search direction
+// at each iteration, and a Linesearcher which performs a linesearch along
```
btracey: I don't like the overloading of the word "iteration". How about "specifies the search direction of each line search, and a Linesearcher which performs a line search along that direction"

btracey:
LGTM with minor comments fixed.

PR #113, opened 2015-08-01
Update BFGS to new VecMul API

btracey:


btracey:
Merging because fixes build. Fails because of beta3.

PTAL @kortschak @vladimir-ch

vladimir-ch:
Retrospective LGTM

PR #114, opened 2015-08-03
Loosen Big5 tolerance

btracey:


btracey:
Merging because fixes build. I think this is just a numeric issue. The BiggsEXP5 tests in other places have a lower tolerance as well.

PTAL @vladimir-ch

vladimir-ch:
Retrospective LGTM

PR #117, opened 2015-08-19
Change BFGS to use vectors instead of matrices.

btracey:
There is now a MulVec. Use it.

### DIFF
--- a/bfgs.go
+++ b/bfgs.go
@@ -151,10 +151,10 @@ func (b *BFGS) NextDirection(loc *Location, dir []float64) (stepSize float64) {
 	copy(b.grad, loc.Gradient)
 
 	// Compute the new search direction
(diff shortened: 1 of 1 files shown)


bfgs.go
```diff
@@ -151,10 +151,10 @@ func (b *BFGS) NextDirection(loc *Location, dir []float64) (stepSize float64) {
 	copy(b.grad, loc.Gradient)
 
 	// Compute the new search direction
-	dirmat := mat64.NewDense(b.dim, 1, dir)
-	gradmat := mat64.NewDense(b.dim, 1, loc.Gradient)
+	dirmat := mat64.NewVector(b.dim, dir)
```
vladimir-ch: Could you also rename {dir,grad}mat to probably just d and g? These variables are very local and not matrices anymore. Then LGTM.

PR #118, opened 2015-08-23
optimize: rename ErrMismatch -> MethodMismatch

kortschak:
I feel this is a more evocative name.

As an aside, it is not used anywhere. Is it actually needed?

@btracey @vladimir-ch Please take a look.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -4,10 +4,7 @@
 
 package optimize
 
-import (
-	"errors"
-	"fmt"
-)
+import "errors"
 
 var (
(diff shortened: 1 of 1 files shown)


btracey:
LGTM. 

I don't think it's needed, because we switched to panicking on method mismatch.

kortschak:
Then delete?

vladimir-ch:
I am for deleting.

btracey:
SGTM

kortschak:
Deleted.

PR #120, opened 2015-09-12
Fix newton with new Cholesky API

btracey:


vladimir-ch:
`resizeTriDense` thus becomes unnecessary, so we should remove it. Other than that, LGTM.

PR #121, opened 2015-09-13
optimize: make chol field concrete type

kortschak:
It is always used, so allocate it with the Newton struct.

@btracey Please take a look.


btracey:
LGTM

Issue #103, opened 2015-05-12
Clearly define meaning of the various iteration types


vladimir-ch:
... and how that effects us maintaining OptLoc and testing for convergence

vladimir-ch:
Continuing here the discussion started in the closely related #93.

Just to sum up the current situation: the core of the API is the reverse-communication interface implemented via `Method.Init()` and `Method.Iterate()`. The idea is that Methods do only method-specific computations and when they need to performs common actions like evaluate the objective function or check for convergence, they return from Iterate() and via xNext, IterationType and EvaluationType communicate to us what they want us to do. We do stuff, fill in Location and call Iterate() again.

The problem is that the iteration types have a broad and not yet clearly defined meaning. Also, since Iterate() returns both IterationType and EvaluationType, there are also many combinations whose meaning we should define, document and deal with. We should also define and document the order in which things happen, e.g., that evaluations (should) take place before anything else.

One possibility to avoid this and make things clearer is for Init() and Iterate() return only the IterationType. Each IterationType would have a much narrower scope or in other words one iteration type would mean one action for us to do. So instead of defining what Minor- and SubIteration mean, I would get rid of them. Morever, IterationType and EvaluationType always appear together (except for Linesearcher), so it makes sense to merge them into one type and thus avoid their possible combinations. It seems that there do not need to be many iteration types: MajorIteration (check convergence and increase iteration count), three evaluation types and (possibly) an iteration type that would instruct us to store the location as an candidate for the next convergence check (especially for Nelder-Mead). ORing of iteration types would not be allowed, except perhaps for the three evaluations. This would keep things simple and clear.

I think that the above would require us to loosen/drop the requirement that Methods cannot modify location, but maybe that is not such a big deal. Location would simply become the one communication point between the client and Method.

The main optimization loop then would become something like (just a sketch):

```go
iterType, err := method.Init(loc, xNext)

for {
	switch iterType {
	case FuncEvaluation, GradEvaluation, HessEvaluation:
		// Perform iterType evaluation of the function at xNext and store the
		// result in location.
		evaluate(p, iterType, xNext, loc, stats) // We could even remove evaluate()
	case MajorIteration:
		stats.MajorIterations++
		status = checkConvergence(optLoc, iterType, stats, settings)
		if status != NotTerminated {
			return
		}
	case StoreCandidate:
		// update optLoc?
	default:
		// Should not happen.
	}

	// Check runtime and evaluations limits, record, check status, ...

[...]

Issue #96, opened 2015-04-15
Statuser docs are not accurate


vladimir-ch:
Also `Method` can implement `Status()` which is checked in `Local()` before calling `Method.Iterate()`.

PR #123, opened 2015-09-14
Merge IterationType and EvaluationType, simplify Method interface

vladimir-ch:
A proposal for resolving #103 ~~but still work in progress~~.

* Merged `IterationType` and `EvaluationType` into ~~`RequestType`~~ `Operation`.
* Simplified `Method` interface. It can't be any simpler than this, so if this PR is accepted, further changes are very unlikely.
* The above two modifications ripple through the code and lead at some places to essentially a complete rewrite, most importantly in `LinesearchMethod` and `minimize()`. But the issue at hand is solved, and the code is now much cleaner and more correct. I am ~~quite~~ very happy with the changes (if nobody praises you, you must praise yourself).
* Updated docs, especially for `Method` and `Operation`, although while at it, I updated the docs also at other places, which, as a by-product, resolves issues #29 and #96.

@kortschak @btracey Please take a look.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -30,15 +30,15 @@ package optimize
 // combine the Evaluation operations with the Iteration operations.
 type Method interface {
 	// Init initializes the method based on the initial data in loc, updates it
-	// and returns the first request.
+	// and returns the first operation to be carried out by the caller.
 	// The initial location must be valid as specified by Needs().
-	Init(loc *Location) (RequestType, error)
+	Init(loc *Location) (Operation, error)
 
 	// Iterate retrieves data from loc, performs one iteration of the method,
-	// updates loc and returns the next request.
+	// updates loc and returns the next operation.
 	// TODO(vladimir-ch): When decided, say something whether the contents of
 	// Location is preserved between calls to Iterate().
-	Iterate(loc *Location) (RequestType, error)
+	Iterate(loc *Location) (Operation, error)
 
 	// Needs specifies information about the objective function needed by the
 	// optimizer beyond just the function value. The information is used
@@ -69,7 +69,7 @@ type Linesearcher interface {
 	// φ(0) and φ'(0), respectively, and step contains the first trial step
 	// length. It returns the type of evaluation to be performed at
 	// x_0 + step * dir_0.
-	Init(value, derivative float64, step float64) RequestType
+	Init(value, derivative float64, step float64) Operation
 
 	// Finished takes in the values of φ and φ' evaluated at the previous step,
 	// and returns whether a sufficiently accurate minimum of φ has been found.
@@ -78,7 +78,7 @@ type Linesearcher interface {
 	// Iterate takes in the values of φ and φ' evaluated at the previous step
 	// and returns the next step size and the type of evaluation to be
 	// performed at x_k + step * dir_k.
-	Iterate(value, derivative float64) (step float64, r RequestType, err error)
+	Iterate(value, derivative float64) (step float64, op Operation, err error)
 }
 
 // NextDirectioner implements a strategy for computing a new line search
@@ -109,5 +109,5 @@ type StepSizer interface {
 // the progress to StdOut or to a log file. A Recorder must not modify any data.
 type Recorder interface {
 	Init() error
-	Record(*Location, RequestType, *Stats) error
+	Record(*Location, Operation, *Stats) error
 }
--- a/interfaces.go
+++ b/interfaces.go
@@ -8,11 +8,11 @@ package optimize
 type Method interface {
 	// Init initializes the method and stores the first location to evaluate
 	// in xNext.
-	Init(loc *Location, xNext []float64) (EvaluationType, IterationType, error)
+	Init(loc *Location) (RequestType, error)
 
 	// Iterate performs one iteration of the method and stores the next
 	// location to evaluate in xNext.
-	Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error)
+	Iterate(loc *Location) (RequestType, error)
 
 	// Needs specifies information about the objective function needed by the
 	// optimizer beyond just the function value. The information is used
@@ -36,7 +36,7 @@ type Linesearcher interface {
 	// φ(0) and φ'(0), respectively, and step contains the first trial step
 	// length. It returns the type of evaluation to be performed at
 	// x_0 + step * dir_0.
-	Init(value, derivative float64, step float64) EvaluationType
+	Init(value, derivative float64, step float64) RequestType
 
 	// Finished takes in the values of φ and φ' evaluated at the previous step,
 	// and returns whether a sufficiently accurate minimum of φ has been found.
@@ -45,7 +45,7 @@ type Linesearcher interface {
 	// Iterate takes in the values of φ and φ' evaluated at the previous step
 	// and returns the next step size and the type of evaluation to be
 	// performed at x_k + step * dir_k.
-	Iterate(value, derivative float64) (step float64, e EvaluationType, err error)
+	Iterate(value, derivative float64) (step float64, r RequestType, err error)
 }
 
 // NextDirectioner implements a strategy for computing a new line search
@@ -84,5 +84,5 @@ type Statuser interface {
 // the progress to StdOut or to a log file. A Recorder must not modify any data.
 type Recorder interface {
 	Init() error
-	Record(*Location, EvaluationType, IterationType, *Stats) error
+	Record(*Location, RequestType, *Stats) error
 }
--- a/interfaces.go
+++ b/interfaces.go
@@ -13,7 +13,7 @@ package optimize
 // data and their representation, and enables automation of common operations
 // like checking for (various types of) convergence and maintaining statistics.
 //
-// A Method can command an Evaluation or a MajorIteration operation.
+// A Method can command an Evaluation, a MajorIteration or NoOperation operations.
 // An evaluation operation is one or more of the Evaluation operations
 // (FuncEvaluation, GradEvaluation, etc.) which can be combined with
 // the bitwise or operator. In an evaluation operation, the requested routines
@@ -25,8 +25,8 @@ package optimize
 // of the optimization (GradientThreshold, etc.) will be checked using this new
 // minimum.
 //
-// A Method must not return InitIteration and PostIteration requests. These are
-// reserved for the clients to be passed to Recorders. A method must also not
+// A Method must not return InitIteration and PostIteration operations. These are
+// reserved for the clients to be passed to Recorders. A Method must also not
 // combine the Evaluation operations with the Iteration operations.
 type Method interface {
 	// Init initializes the method based on the initial data in loc, updates it
--- a/interfaces.go
+++ b/interfaces.go
@@ -5,13 +5,39 @@
 package optimize
 
 // A Method can optimize an objective function.
+//
+// It uses a reverse-communication interface between the optimization method
+// and the caller. Method acts as a client that asks the caller to perform
+// needed operations via RequestType returned from Init and Iterate methods.
+// This provides independence of the optimization algorithm on user-supplied
+// data and their representation, and enables automation of common operations
+// like checking for (various types of) convergence and maintaining statistics.
+//
+// A Method can command an Evaluation or a MajorIteration operation.
+// An evaluation operation is one or more of the Evaluation operations
+// (FuncEvaluation, GradEvaluation, etc.) which can be combined with
+// the bitwise or operator. In an evaluation operation, the requested routines
+// will be evaluated at the point specified in Location.X. The corresponding
+// fields of Location will be filled with the results from the routine and can
+// be retrieved upon the next call to Iterate. Alternatively, a Method can
+// declare a MajorIteration. In a MajorIteration, all values in Location must
+// be valid and consistent, and are interpreted as a new minimum. Convergence
+// of the optimization (GradientThreshold, etc.) will be checked using this new
+// minimum.
+//
+// A Method must not return InitIteration and PostIteration requests. These are
+// reserved for the clients to be passed to Recorders. A method must also not
+// combine the Evaluation operations with the Iteration operations.
 type Method interface {
-	// Init initializes the method and stores the first location to evaluate
-	// in xNext.
+	// Init initializes the method based on the initial data in loc, updates it
+	// and returns the first request.
+	// The initial location must be valid as specified by Needs().
 	Init(loc *Location) (RequestType, error)
 
-	// Iterate performs one iteration of the method and stores the next
-	// location to evaluate in xNext.
+	// Iterate retrieves data from loc, performs one iteration of the method,
+	// updates loc and returns the next request.
+	// TODO(vladimir-ch): When decided, say something whether the contents of
+	// Location is preserved between calls to Iterate().
 	Iterate(loc *Location) (RequestType, error)
 
 	// Needs specifies information about the objective function needed by the
@@ -22,6 +48,13 @@ type Method interface {
 		Gradient bool
 		Hessian  bool
 	}
+}
+
+// Statuser can report the status and any error. It is intended for methods as
+// an additional error reporting mechanism apart from the errors returned from
+// Init() and Iterate().
+type Statuser interface {
+	Status() (Status, error)
 }
 
 // Linesearcher is a type that can perform a line search. It tries to find an
@@ -70,14 +103,6 @@ type NextDirectioner interface {
 type StepSizer interface {
 	Init(loc *Location, dir []float64) float64
 	StepSize(loc *Location, dir []float64) float64
-}
-
-// Statuser returns the status of the Function being optimized. This can be used
-// by the Function to terminate early, for example with an error. The user can
-// use one of the pre-provided Status constants, or may call NewStatus to create
-// a custom Status value.
-type Statuser interface {
-	Status() (Status, error)
 }
 
 // A Recorder can record the progress of the optimization, for example to print
--- a/linesearch.go
+++ b/linesearch.go
@@ -69,10 +69,6 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
 		return ls.initNextLinesearch(loc.X)
 
 	default:
-		if !ls.lastOp.IsEvaluation() {
-			panic("linesearch: unexpected operation")
-		}
-
 		// Store the result of the previously requested evaluation into ls.loc.
 		if ls.lastOp&FuncEvaluation != 0 {
 			ls.loc.F = loc.F
@@ -116,6 +112,9 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
 	step, op, err := ls.Linesearcher.Iterate(ls.loc.F, projGrad)
 	if err != nil {
 		return ls.error(err)
+	}
+	if !op.isEvaluation() {
+		panic("linesearch: Linesearcher returned invalid operation")
 	}
 
 	if step != ls.lastStep {
@@ -167,7 +166,10 @@ func (ls *LinesearchMethod) initNextLinesearch(xNext []float64) (Operation, erro
 		return ls.error(ErrNonNegativeStepDirection)
 	}
 
-	ls.lastOp = ls.Linesearcher.Init(ls.loc.F, projGrad, step)
+	op := ls.Linesearcher.Init(ls.loc.F, projGrad, step)
+	if !op.isEvaluation() {
+		panic("linesearch: Linesearcher returned invalid operation")
+	}
 
 	floats.AddScaledTo(xNext, ls.x, step, ls.dir)
 	if floats.Equal(ls.x, xNext) {
@@ -181,6 +183,7 @@ func (ls *LinesearchMethod) initNextLinesearch(xNext []float64) (Operation, erro
 	copy(ls.loc.X, xNext) // Move ls.loc to the next evaluation point
 	ls.eval = NoOperation // and invalidate all its fields.
 
+	ls.lastOp = op
 	return ls.lastOp, nil
 }
 
--- a/linesearch.go
+++ b/linesearch.go
@@ -64,8 +64,9 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
 
 	case MajorIteration:
 		// We previously requested MajorIteration but since we're here, the
-		// previous location was not good enough to converge. So start the next
-		// linesearch and store the next evaluation point in loc.X.
+		// previous location was not good enough to converge the full
+		// optimization. Start the next linesearch and store the next
+		// evaluation point in loc.X.
 		return ls.initNextLinesearch(loc.X)
 
 	default:
--- a/linesearch.go
+++ b/linesearch.go
@@ -10,14 +10,13 @@ import (
 	"github.com/gonum/floats"
 )
 
-// LinesearchMethod represents an abstract optimization method in which
-// a function is optimized through successive line search optimizations.
-// It consists of a NextDirectioner, which specifies the search direction
-// of each linesearch, and a Linesearcher which performs a linesearch along
-// the search direction.
+// LinesearchMethod represents an abstract optimization method in which a
+// function is optimized through successive line search optimizations.
 type LinesearchMethod struct {
+	// NextDirectioner specifies the search direction of each linesearch.
 	NextDirectioner NextDirectioner
-	Linesearcher    Linesearcher
+	// Linesearcher performs a linesearch along the search direction.
+	Linesearcher Linesearcher
 
 	x   []float64 // Starting point for the current iteration.
 	dir []float64 // Search direction for the current iteration.
--- a/linesearch.go
+++ b/linesearch.go
@@ -19,15 +19,20 @@ type LinesearchMethod struct {
 	NextDirectioner NextDirectioner
 	Linesearcher    Linesearcher
 
-	x     []float64 // Starting point for the current iteration.
-	dir   []float64 // Search direction for the current iteration.
-	first bool      // Indicator of the first iteration.
+	x   []float64 // Starting point for the current iteration.
+	dir []float64 // Search direction for the current iteration.
 
-	evalType EvaluationType
-	iterType IterationType
+	first     bool // Indicator of the first iteration.
+	nextMajor bool // Indicates that MajorIteration must be requested at the next call to Iterate().
+
+	loc  Location    // Storage for intermediate locations.
+	eval RequestType // Indicator of valid fields in loc.
+
+	lastStep    float64     // Step taken from x in the previous call to Iterate().
+	lastRequest RequestType // Request returned from the previous call to Iterate().
 }
 
-func (ls *LinesearchMethod) Init(loc *Location, xNext []float64) (EvaluationType, IterationType, error) {
+func (ls *LinesearchMethod) Init(loc *Location) (RequestType, error) {
 	if loc.Gradient == nil {
 		panic("linesearch: gradient is nil")
 	}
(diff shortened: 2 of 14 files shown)


btracey:
I like RequestType, but maybe it's a little too passive? It's not really a request, as the Method will get back what it wants. Perhaps just "Operation"?

The comments around evaluate are good, but the important part to add the information is on Method.

btracey:
Reading your other comment now.

vladimir-ch:
I like `Operation`, code updated.

I think this is ready for a real review. This PR is quite big, but such a fundamental change cannot be done in small steps. The core changes are in types.go, interfaces.go, local.go and linesearch.go. I will appreciate suggestions for improving the docs. @btracey @kortschak Please take a look.

@btracey Updating NelderMead was easy in the end, the code is well-structured, so it was enough just to add an extra iteration type and a case to a switch. In NelderMead the benefit of this PR is probably most visible. Instead of returning unclear SubIteration and MinorIteration, we simply return FuncEvaluation and when the best value in the simplex could have possibly changed, MajorIteration is returned before proceeding to the reflection step. In both cases the effect is clear even without knowing the context in which the method is used.

local.go
```diff
@@ -13,12 +13,18 @@ import (
 	"github.com/gonum/matrix/mat64"
 )
 
-// Local finds a local minimum of a function using a sequential algorithm.
-// In order to maximize a function, multiply the output by -1.
+// Local finds a local minimum of a minimization problem using a sequential
+// algorithm.
```
btracey: I think it's worth saying something along the lines of "A maximization problem can be transformed into a minimization problem by multiplying the function by -1". A lot of 'commercial' solvers allow users to specify either maximize or minimize, so many users new to optimization don't realize the transformation is that easy.

local.go
```diff
@@ -13,12 +13,18 @@ import (
...
 //
-// The first argument is of Function type representing the function to be minimized.
-// Type switching is used to see if the function implements Gradient, FunctionGradient
-// and Statuser.
+// The first argument represents the problem to be minimized. Its fields are
+// functions that evaluate various quantities related to the problem like the
```
btracey: How about it's fields are "routines"? It's obvious from the type that by routine we mean 'go function', and trying to minimize the number of different meanings of the word "function" is probably our own objective function :)

local.go
```diff
@@ -13,12 +13,18 @@ import (
...
+// objective function or its gradient. p.Func must not be nil and also non-nil
```
btracey: How about "that evaluate the objective function, gradient, and other quantities related to the problem". Push the complexity later in the sentence.

btracey: I would say
"The objective function, p.Func, must not be nil. The optimization method used may require other fields to be non-nil as specified by method.Needs(). Local will panic if these are not met. The method can be determined automatically from the supplied method which will be described later."

local.go
```diff
@@ -124,37 +134,49 @@ func Local(p Problem, initX []float64, settings *Settings, method Method) (*Resu
...
+
+		case InitIteration:
+			panic("optimize: Method returned InitIteration")
+
+		case PostIteration:
+			panic("optimize: Method returned PostIteration")
+
+		case MajorIteration:
+			stats.MajorIterations++
+			if loc.F <= optLoc.F {
```
btracey: I don't think we want to have this check any more. The Method is now in control of how OptLoc is maintained, and so if it wants to have a higher F that's it's prerogative. In particular, with this improved control flow, we could add back in the "floating point noise robustness" code.

local.go
```diff
@@ -124,37 +134,49 @@ func Local(p Problem, initX []float64, settings *Settings, method Method) (*Resu
...
+			status = checkConvergence(optLoc, settings)
 
-		// Perform evalType evaluation of the function at xNext and store the
-		// result in location.
-		evaluate(p, evalType, xNext, loc, stats)
-		// Update the stats and optLoc.
-		update(loc, optLoc, stats, iterType, startTime)
-		// Get the convergence status before recording the new location.
-		status = checkConvergence(optLoc, iterType, stats, settings)
+		default:
```
btracey: Please comment that "default = any of the evaluation cases"

local.go
```diff
@@ -124,37 +134,49 @@ func Local(p Problem, initX []float64, settings *Settings, method Method) (*Resu
...
+		return
 	}
 
 	for {
-		if p.Status != nil {
-			// Check the function status before evaluating.
-			status, err = p.Status()
-			if err != nil || status != NotTerminated {
-				return
+		switch op {
```
btracey: This is a long for loop. Could you just add a quick comment saying something to the effect of "Sequentially call method.Iterate(), performing the operations it has requested, until convergence"

local.go
```diff
@@ -246,60 +272,64 @@ func getStartingLocation(p *Problem, method Method, initX []float64, stats *Stat
...
-	if iterType == MajorIteration || iterType == InitIteration {
-		if loc.Gradient != nil {
-			norm := floats.Norm(loc.Gradient, math.Inf(1))
-			if norm < settings.GradientThreshold {
-				return GradientThreshold
-			}
-		}
-		if loc.F < settings.FunctionThreshold {
-			return FunctionThreshold
+func checkConvergence(loc *Location, settings *Settings) Status {
```
btracey: This code is so simple now!

local.go
```diff
@@ -246,60 +272,64 @@ func getStartingLocation(p *Problem, method Method, initX []float64, stats *Stat
...
+	}
+
+	if settings.FunctionConverge != nil {
+		return settings.FunctionConverge.FunctionConverged(loc.F)
 	}
 
+	return NotTerminated
+}
+
+func checkLimits(loc *Location, stats *Stats, settings *Settings) Status {
```
btracey: Please add a comment that this distinguishing feature between checkLimits and checkConvergence is that checkLimits is checked every evaluation, while checkConvergence is only examined at MajorIteration.

local.go
```diff
@@ -246,60 +272,64 @@ func getStartingLocation(p *Problem, method Method, initX []float64, stats *Stat
...
 	// Check every step for negative infinity because it could break the
 	// linesearches and -inf is the best you can do anyway.
 	if math.IsInf(loc.F, -1) {
 		return FunctionNegativeInfinity
 	}
 
+	if settings.MajorIterations > 0 && stats.MajorIterations >= settings.MajorIterations {
```
btracey: Could we update the other checks to use this format? Sorry.

local.go
```diff
@@ -325,20 +355,18 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
...
 	return NotTerminated
 }
 
-// invalidate marks all fields of Location with NaNs and it is the
-// responsibility of Method to issue such EvaluationTypes that make Location
-// valid again. Invalidation is done internally as a help for implementers to
-// detect silent bugs in Methods using inconsistent Location, e.g., using
-// Gradient without a GradEvaluation request.
-func invalidate(loc *Location) {
+// evaluate evaluates the problem p at loc.X using the evaluation eval,
```
btracey: evaluate evaluates the routines specified by the Operation at loc.X, storing the answer into loc and updating stats.

types.go
```diff
@@ -15,54 +15,62 @@ import (
...
+	PostIteration
+	// FuncEvaluation is the request to evaluate the objective function.
+	FuncEvaluation
+	// GradEvaluation is the request to evaluate the gradient of the objective function.
 	GradEvaluation
+	// HessEvaluation is the request to evaluate the Hessian of the objective function.
 	HessEvaluation
+
+	// Mask for the evaluating operations.
+	Evaluations = FuncEvaluation | GradEvaluation | HessEvaluation
```
btracey: Does this need to be exported?

types.go
```diff
@@ -15,54 +15,62 @@ import (
...
 )
 
-func (e EvaluationType) String() string {
-	return fmt.Sprintf("EvaluationType(Func: %t, Grad: %t, Hess: %t, Extra: 0b%b)",
-		e&FuncEvaluation != 0,
-		e&GradEvaluation != 0,
-		e&HessEvaluation != 0,
-		e&^(FuncEvaluation|GradEvaluation|HessEvaluation))
+func (op Operation) IsEvaluation() bool {
+	return op&Evaluations != 0 && op&^Evaluations == 0
```
btracey: @kortschak could you confirm you think this line is correct?

types.go
```diff
@@ -15,54 +15,62 @@ import (
...
+	Evaluations = FuncEvaluation | GradEvaluation | HessEvaluation
 )
 
-func (e EvaluationType) String() string {
-	return fmt.Sprintf("EvaluationType(Func: %t, Grad: %t, Hess: %t, Extra: 0b%b)",
-		e&FuncEvaluation != 0,
-		e&GradEvaluation != 0,
-		e&HessEvaluation != 0,
-		e&^(FuncEvaluation|GradEvaluation|HessEvaluation))
+func (op Operation) IsEvaluation() bool {
```
btracey: Does this need to be exported?

local.go
```diff
@@ -325,20 +355,18 @@ func checkConvergence(loc *Location, iterType IterationType, stats *Stats, setti
...
+// stores the answer into loc and updates stats. Unused fields of loc are set
+// to NaN. It is the responsibility of Method to assemble a valid Location
+// before requesting MajorIteration.
+func evaluate(p *Problem, loc *Location, eval Operation, stats *Stats) {
+	if !eval.IsEvaluation() {
```
btracey: I think this should be in the minimize loop, just after the default case. Check if it's a valid evaluation, and if so evaluate.

types.go
```diff
@@ -15,54 +15,62 @@ import (
 
 const defaultGradientAbsTol = 1e-6
 
-// EvaluationType is used by a Method to specify the objective-function
-// information needed at an x location.
-type EvaluationType uint
+// Operation represents the set of operations requested by Method at each
+// iteration. It is a bitmap of *Iteration and *Evaluation constants.
```
btracey: That notation is confusing. I originally read it as "Pointer to Iteration type" and "Pointer to evaluation type". 

We need to comment that a Method may return MajorIteration xor NoOperation xor a combination of the evaluations. A method may never return Init and Post. It's probably commented in Method, but I don't think it hurts to have it in multiple places in this case.

types.go
```diff
@@ -15,54 +15,62 @@ import (
...
+type Operation uint64
 
-// Evaluation types can be composed together using the binary or operator, for
-// example 'FuncEvaluation | GradEvaluation' to evaluate both the function
-// value and the gradient.
+// Supported Operations. Individual operations must NOT be combined together by
```
btracey: Oh, I see. The rules should be in the Operation comment where they will be visible through godoc.

types.go
```diff
@@ -15,54 +15,62 @@ import (
...
+// the binary OR operator except for the various *Evaluations.
 const (
-	NoEvaluation   EvaluationType = 0
-	FuncEvaluation EvaluationType = 1 << (iota - 1)
+	// NoOperation does not specify any operation.
```
btracey: NoOperation specifies that no evaluation or iteration should take place.

types.go
```diff
@@ -15,54 +15,62 @@ import (
...
+	NoOperation Operation = 0
+	// InitIteration is sent to Recorder to indicate the initial location. All
```
btracey: It's also a signal to methods. "InitIteration is a signal for the initial iteration of the optimization routine. It is sent ..."

types.go
```diff
@@ -15,54 +15,62 @@ import (
...
+	InitIteration Operation = 1 << (iota - 1)
+	// MajorIteration indicates that a Method has found the next candidate
+	// location for an optimum and convergence should be checked.
+	MajorIteration
+	// PostIteration is sent to Recorder to indicate the final location reached
+	// during an optimization run. All fields of the location to record must be
+	// valid.
+	// Methods must not return it.
+	PostIteration
+	// FuncEvaluation is the request to evaluate the objective function.
```
btracey: Again, I don't think request is the right word.

types.go
```diff
@@ -111,14 +119,19 @@ type Problem struct {
 	// Func evaluates the objective function at the given location. Func
 	// must not modify x.
 	Func func(x []float64) float64
+
```
btracey: Not to change in this PR, but I wonder if we really want it to be the rule that x isn't modified, and instead just pass a copy of x. The copy should be reasonably cheap and then there are no bugs.

types.go
```diff
@@ -111,14 +119,19 @@ type Problem struct {
...
 	// Grad evaluates the gradient at x and stores the result in-place in grad.
 	// Grad must not modify x.
 	Grad func(x []float64, grad []float64)
+
```
btracey: Also not for this PR, but I think we should change the input order. The gonum rule is that "answer receivers" are the first inputs.

interfaces.go
```diff
@@ -5,14 +5,31 @@
 package optimize
 
 // A Method can optimize an objective function.
+//
+// It uses a reverse-communication interface design represented by the Init()
+// and Iterate() methods to pass and retrieve information from the client. Its
```
btracey: "methods to receive information about the objective function". It doesn't pass information to the client aside from X location.

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
+// aim is to isolate the method from user-supplied information and data and
```
btracey: How about just "This enables automation of common operations like ...". I'm not sure the "isolate" part of the comment is helpful.

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
+// their representation, and from common operations like checking for (various
+// types of) convergence and maintaining statistics.
+//
+// When Method needs the client to perform an operation like function
```
btracey: I would say something like "A Method can command an Evaluation or a MajorIteration operation. An evaluation operation is one or more of the Evaluaton operations (FunctionEvaluation, GradientEvaluation, etc.) which can be combined with the bitwise or operator. In an evaluation operation, the requested routines will be evaluated at the value specified in Location.X. The corresponding fields of Location will be filled with the results from the routine. Alternately, a method can declare a MajorIteration. In a MajorIteration, the values in Location are interpreted as a new minimum. Convergence of the optimization (GradientThreshold, etc.) will be checked using this new minimum."

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
+// evaluation or check for convergence, it fills the Location with data and
+// returns a corresponding Operation. The result then can be retrieved from
+// Location upon the next call to Iterate().
+//
+// Methods must not return InitIteration and PostIteration operations. These
+// are reserved for the clients to be passed to Recorders.
```
btracey: A method must also not combine the Evaluation operations with the iteration operations.

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
 type Method interface {
-	// Init initializes the method and stores the first location to evaluate
-	// in xNext.
-	Init(loc *Location, xNext []float64) (EvaluationType, IterationType, error)
+	// Init initializes the method based on the initial data in loc, updates it
```
btracey: updates the X location in Loc.

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
+	// and returns the first operation to be carried out by the caller.
+	// The initial location must be valid as specified by Needs().
+	Init(loc *Location) (Operation, error)
 
-	// Iterate performs one iteration of the method and stores the next
-	// location to evaluate in xNext.
-	Iterate(loc *Location, xNext []float64) (EvaluationType, IterationType, error)
+	// Iterate retrieves data from loc, performs one iteration of the method,
+	// updates loc and returns the next operation.
```
btracey: the x location of loc.

btracey:
Did a big chuck of it. Haven't gotten around to the new Linesearcher code or any of the tweaks to the actual Methods.

interfaces.go
kortschak: s/Alternately/Alternatively/

Semantically different; alternate is to swap between two state successively, alternative is to choose another - except in NA.

local.go
```diff
@@ -124,37 +134,49 @@ func Local(p Problem, initX []float64, settings *Settings, method Method) (*Resu
...
+				if err != nil || status != NotTerminated {
+					return
+				}
+			}
+			evaluate(p, loc, op, stats)
+		}
 
 		if settings.Recorder != nil {
-			err = settings.Recorder.Record(loc, evalType, iterType, stats)
+			stats.Runtime = time.Since(startTime)
```
vladimir-ch: I wonder if we shouldn't update Runtime only once per iteration so that the Recorder and checkLimits() get the same value. The complete opposite would be replacing Runtime with StartTime so that the Recorder and checkLimits() are forced to compute the actual runtime at the moment when they need it.

linesearch.go
```diff
@@ -10,24 +10,28 @@ import (
...
-	x     []float64 // Starting point for the current iteration.
-	dir   []float64 // Search direction for the current iteration.
-	first bool      // Indicator of the first iteration.
+	x   []float64 // Starting point for the current iteration.
+	dir []float64 // Search direction for the current iteration.
 
-	evalType EvaluationType
-	iterType IterationType
+	first     bool // Indicator of the first iteration.
+	nextMajor bool // Indicates that MajorIteration must be requested at the next call to Iterate().
```
kortschak: Drop parens.

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
+	// TODO(vladimir-ch): When decided, say something whether the contents of
+	// Location is preserved between calls to Iterate().
```
kortschak: Drop parens.

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
+//
+// Methods must not return InitIteration and PostIteration operations. These
+// are reserved for the clients to be passed to Recorders.
 type Method interface {
-	// Init initializes the method and stores the first location to evaluate
-	// in xNext.
-	Init(loc *Location, xNext []float64) (EvaluationType, IterationType, error)
+	// Init initializes the method based on the initial data in loc, updates it
+	// and returns the first operation to be carried out by the caller.
+	// The initial location must be valid as specified by Needs().
```
kortschak: Drop parens.

interfaces.go
```diff
@@ -5,14 +5,31 @@
...
+// It uses a reverse-communication interface design represented by the Init()
+// and Iterate() methods to pass and retrieve information from the client. Its
+// aim is to isolate the method from user-supplied information and data and
+// their representation, and from common operations like checking for (various
+// types of) convergence and maintaining statistics.
+//
+// When Method needs the client to perform an operation like function
+// evaluation or check for convergence, it fills the Location with data and
+// returns a corresponding Operation. The result then can be retrieved from
+// Location upon the next call to Iterate().
```
kortschak: Drop parens.

interfaces.go
```diff
@@ -5,14 +5,31 @@
 package optimize
 
 // A Method can optimize an objective function.
+//
+// It uses a reverse-communication interface design represented by the Init()
```
kortschak: Drop parens.

interfaces.go
```diff
@@ -24,6 +41,13 @@ type Method interface {
 	}
 }
 
+// Statuser can report the status and any error. It is intended for methods as
+// an additional error reporting mechanism apart from the errors returned from
+// Init() and Iterate().
```
kortschak: Drop parens.

linesearch.go
```diff
@@ -10,24 +10,28 @@ import (
...
+
+	loc  Location  // Storage for intermediate locations.
+	eval Operation // Indicator of valid fields in loc.
+
+	lastStep float64   // Step taken from x in the previous call to Iterate().
```
kortschak: Parens.

linesearch.go
```diff
@@ -35,96 +39,149 @@ func (ls *LinesearchMethod) Init(loc *Location, xNext []float64) (EvaluationType
...
-		copy(xNext, loc.X)
-		ls.evalType = NoEvaluation
-		ls.iterType = MajorIteration
-		return ls.evalType, ls.iterType, nil
-	}
+func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
+	switch ls.lastOp {
+	case NoOperation:
+		// TODO(vladimir-ch): We have previously returned with an error and
+		// Init() was not called. What to do? What about ls's internal state?
```
kortschak: Parens.

local.go
```diff
@@ -27,7 +33,7 @@ import (
 // The third argument contains the settings for the minimization. It is here that
 // gradient tolerance, etc. are specified. The DefaultSettings() function
```
kortschak: Drop parens.

local.go
```diff
@@ -36,18 +42,22 @@ import (
...
-//  method := &Bfgs{}
-// creates a pointer to a new Bfgs struct. When minimize is called, the settings
+//  method := &optimize.BFGS{}
+// creates a pointer to a new BFGS struct. When Local is called, the settings
 // in the method will be populated with default values. The methods are also
-// designed such that they can be reused in future calls to method.
+// designed such that they can be reused in future calls to Local.
+//
+// If method implements Statuser, method.Status() is called before every call
+// to method.Iterate(). If the returned Status is not NotTerminated or the
```
kortschak: Parens.

neldermead.go
```diff
@@ -142,10 +143,9 @@ func (n *NelderMead) Init(loc *Location, xNext []float64) (EvaluationType, Itera
 	copy(n.vertices[dim], loc.X)
 	n.values[dim] = loc.F
 	n.fillIdx = 0
-	copy(xNext, loc.X)
-	xNext[0] += n.SimplexSize
+	loc.X[n.fillIdx] += n.SimplexSize
 	n.lastIter = nmInitialize
-	return FuncEvaluation, InitIteration, nil
+	return FuncEvaluation, nil // InitIteration
```
kortschak: Flesh out this comment or remove it.

types.go
kortschak: Put the Operation and EvaluationType constants in separate const blocks so godoc can render them with their type.

types.go
kortschak: If it is unexported, call it something like evalMask so it's clear in the code what it is doing.

types.go
kortschak: It is.

kortschak:
I'm reviewing from a position of ignorance in most of this, so I focussed on wording and code structure. I should know more, but I don't - it would be nice when the API has stabilised to have some wiki pages outlining use cases (examples are probably likely too long to include, but if you think you can do that, that would be good too).

LGTM FWIW

types.go
vladimir-ch: I am not sure about the word "iteration" in this phrase. No sane Method has a reason to return NoOperation from Iterate(), but I can imagine that maybe perhaps it could want to return NoOperation from Init() if it want to just pass the control to Iterate(). At any case, NoOperation will lead to Iterate() being called again without any particular action apart from checking the limits and sending the loc to Recorder.

types.go
vladimir-ch: We don't send Operations to Methods. Previously NelderMead returned InitIteration for the evaluation of the initial simplex, but I think the new framework is cleaner and more consistent. Evaluations of the initial simplex are simply evaluations and when it has been evaluated, MajorIteration is returned because a satisfactory minimum might have been found. So there is no place for returning InitIteration either.

types.go
vladimir-ch: Good idea. There is always someone who wants to modify x with a good, but misguided intention.

types.go
vladimir-ch: Also agreed.

interfaces.go
vladimir-ch: It does when MajorIteration is returned. In that case the whole Location might be filled by Method.

I am going to rewrite this paragraph anyway.

interfaces.go
vladimir-ch: Like I wrote above, Method can set also other fields of Loc when it is asking for MajorIteration. That's why I didn't write just X.

local.go

local.go

local.go

local.go

local.go

local.go

local.go
vladimir-ch: Yes! And also evaluate and LinesearchMethod ...

local.go

local.go

local.go

types.go
vladimir-ch: Unexported and renamed to evalMask.

types.go
vladimir-ch: Unexported.

local.go

types.go
vladimir-ch: I thought that using * glob was clever, but I did not realize the connotation with pointer. I removed the * and reorganized the docs a bit.

types.go
vladimir-ch: I moved the rules to the Operation comment, but I think they would be visible through godoc even if they stayed here.

types.go
vladimir-ch: I went with "NoOperation specifies that no evaluation or convergence check should take place."

types.go
vladimir-ch: I changed it to "FuncEvaluation specifies that the objective function should be evaluated." I think using "should" is ok, because Problem.Status can return bad status, so the evaluation might not take place or Iterate() might not be called again.

interfaces.go
vladimir-ch: I changed it to:

```
// This provides independence of the optimization algorithm on user-supplied
// data and their representation, and enables automation of common operations
// like checking for (various types of) convergence and maintaining statistics.
```

I wanted to keep the mention of user data and their representation. With Local() we make and fix all the choices for the user, but there is still the freedom to take a Method and use it in a custom driver routine with different representation of the problem, settings, convergence check, anything.

interfaces.go
vladimir-ch: I went with your version.

interfaces.go
vladimir-ch: Done.

interfaces.go
vladimir-ch: I kept the comment as it is for the moment.

neldermead.go
vladimir-ch: Ooops, a leftover. Removed.

vladimir-ch:
@kortschak Thanks for taking a look! I will postpone dropping the parenthesis to another PR. It occurs also at other places in the package, so I will remove them all in one pass later.

@btracey I have addressed your comments. Mostly I took your suggestions for the documentation literally, it was really helpful, they are great. You should look again at what I wrote for Method, especially the first paragraph, and for Operation. After this the main changes to be reviewed are in LinesearchMethod (major changes) and NelderMead (minor changes). The rest is just noise coming from the merge IterationType+EvaluationType => Operation and from the modified Method interface.

linesearch.go
```diff
@@ -10,24 +10,28 @@ import (
...
-	first bool      // Indicator of the first iteration.
+	x   []float64 // Starting point for the current iteration.
+	dir []float64 // Search direction for the current iteration.
 
-	evalType EvaluationType
-	iterType IterationType
+	first     bool // Indicator of the first iteration.
+	nextMajor bool // Indicates that MajorIteration must be requested at the next call to Iterate().
+
+	loc  Location  // Storage for intermediate locations.
```
vladimir-ch: When you get here, @btracey , this is the key point in the discussion about preserving the contents of Location. If we are ok with saying that LinesearchMethod must receive the non-evaluated fields of Location unchanged, then this field can go away and we save (in some cases many) unnecessary copies of Gradient (and Hessian in case of Newton). Instead, we could build the location gradually in the Location passed to Init and Iterate.

For example conjugate gradient produces badly scaled directions and the line searches rarely finish in one iteration. Every point that does not satisfy Armijo or Wolfe means an extra copy. For expensive functions probably not significant, but still ...

linesearch.go
```diff
@@ -35,96 +39,152 @@ func (ls *LinesearchMethod) Init(loc *Location, xNext []float64) (EvaluationType
...
+		// linesearch and store the next evaluation point in loc.X.
+		return ls.initNextLinesearch(loc.X)
+
+	default:
+		// Store the result of the previously requested evaluation into ls.loc.
+		if ls.lastOp&FuncEvaluation != 0 {
+			ls.loc.F = loc.F
+		}
+		if ls.lastOp&GradEvaluation != 0 {
+			copy(ls.loc.Gradient, loc.Gradient)
```
vladimir-ch: This ...

linesearch.go
```diff
@@ -35,96 +39,152 @@ func (ls *LinesearchMethod) Init(loc *Location, xNext []float64) (EvaluationType
...
+		}
+		if ls.lastOp&HessEvaluation != 0 {
+			ls.loc.Hessian.CopySym(loc.Hessian)
```
vladimir-ch: ... and this are the copies that could go away. Especially this copy will be the most wasteful. Newton will often finish the linesearch with just one iteration, then HessEvaluation will be returned, then here a copy to ls.loc and then few line below a copy of the whole location back to loc.

linesearch.go
btracey: I'm okay with keeping the values in Linesearch. I would like for it to happen in the next PR as this one is already huge (I understand the reason, but we don't need to make it bigger).

interfaces.go
```diff
@@ -5,14 +5,40 @@
...
 // A Method can optimize an objective function.
+//
+// It uses a reverse-communication interface between the optimization method
+// and the caller. Method acts as a client that asks the caller to perform
+// needed operations via RequestType returned from Init and Iterate methods.
+// This provides independence of the optimization algorithm on user-supplied
+// data and their representation, and enables automation of common operations
+// like checking for (various types of) convergence and maintaining statistics.
+//
+// A Method can command an Evaluation or a MajorIteration operation.
```
btracey: We need to add NoIteration

linesearch.go
btracey: Init() sets last op to NoOp, so we can't tell if init was called or not based on this state.

linesearch.go
```diff
@@ -35,96 +39,152 @@ func (ls *LinesearchMethod) Init(loc *Location, xNext []float64) (EvaluationType
...
-	}
+func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
+	switch ls.lastOp {
+	case NoOperation:
+		// TODO(vladimir-ch): We have previously returned with an error and
+		// Init() was not called. What to do? What about ls's internal state?
+
+	case MajorIteration:
+		// We previously requested MajorIteration but since we're here, the
+		// previous location was not good enough to converge. So start the next
```
btracey: Add "converge the full optimization", just so there's no confusion with "converge the line search". Remove the "So" (don't pick that up from me :) )

linesearch.go
```diff
@@ -35,96 +39,152 @@ func (ls *LinesearchMethod) Init(loc *Location, xNext []float64) (EvaluationType
...
 	}
 
-	floats.AddScaledTo(xNext, ls.x, stepSize, ls.dir)
-	// Compare the starting point for the current iteration with the next
-	// evaluation point to make sure that rounding errors do not prevent progress.
-	if floats.Equal(ls.x, xNext) {
-		ls.evalType = NoEvaluation
-		ls.iterType = NoIteration
-		return ls.evalType, ls.iterType, ErrNoProgress
+	if step != ls.lastStep {
```
btracey: Would you mind swapping the if to have the equals case first? My first thought is "what happens if they're equal" and want to read below. Plus, that case is simpler and easy to digest.

btracey:
LGTM I think with the few minor cleanups.

linesearch.go
vladimir-ch: `initNextLinesearch()` then overwrites `lastOp` with the operation returned from `Linesearcher` (which must not return `NoOperation`) so theoretically we could, but probably we should not. We will still have to decide what is the proper thing to do after an error return. I will modify the TODO text and file an issue so that we don't forget.

btracey: Yes, please. The tests are passing now, and we'll be changing anyway with
the copying issue. This way we can think of the algorithm on the whole
without am the extra changes

interfaces.go
vladimir-ch: Done

linesearch.go
vladimir-ch: Done. Done (I won't :) ).

linesearch.go
vladimir-ch: Done

btracey:
Thanks! This is a very nice fix to the major problem in the package design.

types.go
kortschak: Sometimes I don't understand the things that I write. I'll put this one down to temporary colour blindness.

vladimir-ch: Ah, you noticed @kortschak . I kept a tactful silence :-)

kortschak: I noticed that I say foolish things long ago.

You can see that in my github comment history going a fair way back.

PR #131, opened 2015-09-24
Stop modifying Location beyond requested evaluations

vladimir-ch:
* Stop invalidating in `evaluate()`
* Remove `LinesearchMethod.loc` and instead gradually build up a valid location from the passed `Location`. This avoids a lot of unnecessary copying and results in simpler code. It is a horribly crude benchmark, but on my machine `go test` finishes regularly in 7.8-7.9 s with this change which is roughly 0.5 s faster than before.

@btracey Please take a look

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -8,22 +8,28 @@ package optimize
 //
 // It uses a reverse-communication interface between the optimization method
 // and the caller. Method acts as a client that asks the caller to perform
-// needed operations via RequestType returned from Init and Iterate methods.
+// needed operations via Operation returned from Init and Iterate methods.
 // This provides independence of the optimization algorithm on user-supplied
 // data and their representation, and enables automation of common operations
 // like checking for (various types of) convergence and maintaining statistics.
 //
 // A Method can command an Evaluation, a MajorIteration or NoOperation operations.
+//
 // An evaluation operation is one or more of the Evaluation operations
 // (FuncEvaluation, GradEvaluation, etc.) which can be combined with
-// the bitwise or operator. In an evaluation operation, the requested routines
-// will be evaluated at the point specified in Location.X. The corresponding
-// fields of Location will be filled with the results from the routine and can
-// be retrieved upon the next call to Iterate. Alternatively, a Method can
-// declare a MajorIteration. In a MajorIteration, all values in Location must
-// be valid and consistent, and are interpreted as a new minimum. Convergence
-// of the optimization (GradientThreshold, etc.) will be checked using this new
-// minimum.
+// the bitwise or operator. In an evaluation operation, the requested fields of
+// Problem will be evaluated at the point specified in Location.X.
+// The corresponding fields of Location will be filled with the results that
+// can be retrieved upon the next call to Iterate. The Method interface
+// requires that entries of Location are not modified aside from the commanded
+// evaluations. Thus, the type implementing Method may use multiple Operations
+// to set the Location fields at a particular x value.
+//
+// Instead of an Evaluation, a Method may declare MajorIteration. In
+// a MajorIteration, the values in the fields of Location are treated as
+// a potential optimizer. The convergence of the optimization routine
+// (GradientThreshold, etc.) is checked at this new best point. In
+// a MajorIteration, the fields of Location must be valid and consistent.
 //
 // A Method must not return InitIteration and PostIteration operations. These are
 // reserved for the clients to be passed to Recorders. A Method must also not
@@ -36,8 +42,6 @@ type Method interface {
 
 	// Iterate retrieves data from loc, performs one iteration of the method,
 	// updates loc and returns the next operation.
-	// TODO(vladimir-ch): When decided, say something whether the contents of
-	// Location is preserved between calls to Iterate.
 	Iterate(loc *Location) (Operation, error)
 
 	// Needs specifies information about the objective function needed by the
--- a/linesearch.go
+++ b/linesearch.go
@@ -22,7 +22,7 @@ type LinesearchMethod struct {
 	dir []float64 // Search direction for the current iteration.
 
 	first     bool      // Indicator of the first iteration.
-	nextMajor bool      // Indicates that MajorIteration must be requested at the next call to Iterate.
+	nextMajor bool      // Indicates that MajorIteration must be commanded at the next call to Iterate.
 	eval      Operation // Indicator of valid fields in Location.
 
 	lastStep float64   // Step taken from x in the previous call to Iterate.
--- a/linesearch.go
+++ b/linesearch.go
@@ -21,11 +21,9 @@ type LinesearchMethod struct {
 	x   []float64 // Starting point for the current iteration.
 	dir []float64 // Search direction for the current iteration.
 
-	first     bool // Indicator of the first iteration.
-	nextMajor bool // Indicates that MajorIteration must be requested at the next call to Iterate.
-
-	loc  Location  // Storage for intermediate locations.
-	eval Operation // Indicator of valid fields in loc.
+	first     bool      // Indicator of the first iteration.
+	nextMajor bool      // Indicates that MajorIteration must be requested at the next call to Iterate.
+	eval      Operation // Indicator of valid fields in Location.
 
 	lastStep float64   // Step taken from x in the previous call to Iterate.
 	lastOp   Operation // Operation returned from the previous call to Iterate.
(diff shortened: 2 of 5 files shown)


interfaces.go
```diff
@@ -8,22 +8,27 @@ package optimize
...
 // data and their representation, and enables automation of common operations
 // like checking for (various types of) convergence and maintaining statistics.
 //
 // A Method can command an Evaluation, a MajorIteration or NoOperation operations.
+//
 // An evaluation operation is one or more of the Evaluation operations
 // (FuncEvaluation, GradEvaluation, etc.) which can be combined with
 // the bitwise or operator. In an evaluation operation, the requested routines
 // will be evaluated at the point specified in Location.X. The corresponding
 // fields of Location will be filled with the results from the routine and can
```
btracey: "fields of Problem" instead of routine? Not sure.

interfaces.go
```diff
@@ -8,22 +8,27 @@ package optimize
...
-// be retrieved upon the next call to Iterate. Alternatively, a Method can
-// declare a MajorIteration. In a MajorIteration, all values in Location must
-// be valid and consistent, and are interpreted as a new minimum. Convergence
-// of the optimization (GradientThreshold, etc.) will be checked using this new
-// minimum.
+// be retrieved upon the next call to Iterate. The caller must not modify
```
btracey: I see what you're getting at, but I don't think this is phrased properly. The usual case for implementing Method is to use with Local. Maybe something like "The Method interface requires that entries of Location are not modified aside from the commanded evaluations. Thus, the type implementing Method may use multiple Operations to set the Location fields at a particular x value."

interfaces.go
```diff
@@ -8,22 +8,27 @@ package optimize
...
+// the contents of Location between calls to Init and Iterate apart from the
+// commanded evaluations so that a Method can use it as a storage for gradually
+// building a valid location.
+//
+// Alternatively, a Method can declare a MajorIteration. In a MajorIteration,
```
btracey: Instead of an Evaluation type, a method may declare MajorIteration. In a MajorIteration, the values in the fields of Location are treated as a potential optimizer. The convergence of the optimization routine (GradientThreshold, etc.) is checked at this new best point. In a MajorIteration, the fields of Location must be valid and consistent.

linesearch.go
```diff
@@ -181,8 +161,7 @@ func (ls *LinesearchMethod) initNextLinesearch(xNext []float64) (Operation, erro
 	}
 
 	ls.lastStep = step
-	copy(ls.loc.X, xNext) // Move ls.loc to the next evaluation point
-	ls.eval = NoOperation // and invalidate all its fields.
+	ls.eval = NoOperation // Invalidate all fields of loc.
```
btracey: "Indicate all invalid fields of loc" to mirror the comment in Init.

linesearch.go
```diff
@@ -149,31 +129,31 @@ func (ls *LinesearchMethod) error(err error) (Operation, error) {
...
 
 // initNextLinesearch initializes the next linesearch using the previous
-// complete location stored in ls.loc. It fills xNext and returns an evaluation
-// to be performed at xNext.
-func (ls *LinesearchMethod) initNextLinesearch(xNext []float64) (Operation, error) {
-	copy(ls.x, ls.loc.X)
+// complete location stored in loc. It fills loc.X and returns an evaluation
+// to be performed at loc.X.
+func (ls *LinesearchMethod) initNextLinesearch(loc *Location) (Operation, error) {
+	copy(ls.x, loc.X)
```
btracey: Do we need to set ls.lastOp to be NoOperation up here so that's what it equals if an error occurs?

linesearch.go
```diff
@@ -63,46 +60,34 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
 		// Init was not called. What to do? What about ls's internal state?
```
btracey: I still don't understand this comment. What does "and init was not called" mean? Do you mean "or init was not called"? It seems like we can get in this state in some error cases or if init was not called. I think the right thing is to document that if an error was returned the LinesearchMethod must be re-initialized. For example, if the linesearch fails the Method may try again in a different direction. This should be fine, just Init needs to be called again (with a valid location).

btracey: Presumably this can then never happen, and could panic, unless we think there is a valid reason for a Linesearcher to return NoOperation.

linesearch.go
```diff
@@ -63,46 +60,34 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
...
 
 	case MajorIteration:
-		// We previously requested MajorIteration but since we're here, the
+		// We previously commanded MajorIteration but since we're here, the
```
btracey: "The previous updated location did not converge the full optimization. Initialize a new Linesearch."

linesearch.go
```diff
@@ -63,46 +60,34 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
...
-		// Update the indicator of valid fields of ls.loc.
+		// Update the indicator of valid fields of loc.
 		ls.eval |= ls.lastOp
 
 		if ls.nextMajor {
 			ls.nextMajor = false
 
 			// Linesearcher previously indicated that it had finished, but we
-			// needed to evaluate invalid fields of ls.loc. Now we have them and
+			// needed to evaluate invalid fields of loc. Now we have them and
```
btracey: Linesearcher previously finished, and the invalid fields of loc have now been validated. Announce MajorIteration.

linesearch.go
```diff
@@ -63,46 +60,34 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
...
-			ls.loc.F = loc.F
-		}
-		if ls.lastOp&GradEvaluation != 0 {
-			copy(ls.loc.Gradient, loc.Gradient)
-		}
-		if ls.lastOp&HessEvaluation != 0 {
-			ls.loc.Hessian.CopySym(loc.Hessian)
-		}
-		// Update the indicator of valid fields of ls.loc.
+		// Update the indicator of valid fields of loc.
```
btracey: Move the indicator below the nextMajor check. It's not needed there, so it obfuscates.

linesearch.go
```diff
@@ -63,46 +60,34 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
...
 			ls.lastOp = MajorIteration
 			return ls.lastOp, nil
 		}
 	}
 
-	projGrad := floats.Dot(ls.loc.Gradient, ls.dir)
-	if ls.Linesearcher.Finished(ls.loc.F, projGrad) {
-		// Form an operation that evaluates invalid fields of ls.loc.
-		ls.lastOp = complementEval(&ls.loc, ls.eval)
+	projGrad := floats.Dot(loc.Gradient, ls.dir)
```
btracey: This is wrong if loc.Gradient isn't valid. If not valid set projGrad to NaN?

linesearch.go
```diff
@@ -63,46 +60,34 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
...
+			// needed to evaluate invalid fields of loc. Now we have them and
 			// can announce MajorIteration.
 
-			copyLocation(loc, &ls.loc)
 			ls.lastOp = MajorIteration
 			return ls.lastOp, nil
 		}
 	}
 
-	projGrad := floats.Dot(ls.loc.Gradient, ls.dir)
```
btracey: Could you add a comment here like "Continue the linesearch" just to confirm to the code reader that this is the main line? (followed by the update valid fields)

linesearch.go
```diff
@@ -118,11 +103,7 @@ func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
 		panic("linesearch: Linesearcher returned invalid operation")
 	}
 
-	if step == ls.lastStep {
-		// Linesearcher is requesting another evaluation at the same point
-		// which is stored in ls.loc.X.
-		copy(loc.X, ls.loc.X)
-	} else {
+	if step != ls.lastStep {
 		// We are moving to a new location.
```
btracey: "We are moving to a new location, and not, say, evaluating extra information at the current location."

btracey:
I hope you don't mind I took the liberty to take a second pass at a lot of the existing comments. Otherwise, two seemingly small issues. LGTM assuming the issues aren't trickier than I think.

linesearch.go
vladimir-ch: On error I call `LinesearchMethod.error()` which sets lastOp to NoOperation.

btracey: Oh, sorry, I missed that. Thanks.

linesearch.go
vladimir-ch: Calling Iterate() without calling Init() could indeed panic, but in this case I was considering the situation when Iterate() returned with an error and the caller is trying to resume the optimization after doing something about the error without calling Init() (because that is what we discussed previously). `Local` precludes this use case and always calls Init(), but I still operate under the assumption that Local is a kind of convenience wrapper around the iteration loop and someone might want more flexibility and use its own loop.

Calling Init() with the location at the last MajorIteration seems to be the best universal solution but I had the impression that you did not want to enforce it. With the previous big PR, this one and with the changes related to removing Finished() from Linesearcher that I have in my mind I could not sort out my thoughts on this.

linesearch.go
vladimir-ch: Although not necessary, I would like `ls.eval` to be complete when calling `initNextLinesearch` and for that it has to stay here.

linesearch.go
vladimir-ch: I planned to update this part properly when removing Finished from Linesearcher but I will include it here.

vladimir-ch:
> I hope you don't mind I took the liberty to take a second pass at a lot of the existing comments.

Not at all, quite the opposite. In fact, could you give it another, quick pass? I mostly went with your wording, but I also added the NaNing of f and projGrad given to the Linesearcher.

linesearch.go
```diff
@@ -43,66 +41,60 @@ func (ls *LinesearchMethod) Init(loc *Location) (Operation, error) {
...
-	return ls.initNextLinesearch(loc.X)
+	return ls.initNextLinesearch(loc)
 }
 
 func (ls *LinesearchMethod) Iterate(loc *Location) (Operation, error) {
 	switch ls.lastOp {
 	case NoOperation:
-		// TODO(vladimir-ch): We have previously returned with an error and
-		// Init was not called. What to do? What about ls's internal state?
+		// TODO(vladimir-ch): Either Init has not been called, or the caller is
```
btracey: Sounds good, we'll leave to another PR for fix.

btracey:
LGTM.

PR #132, opened 2015-09-24
Rename some error variables and update their comments

### LINKED ISSUE #124: Better name for ErrNonNegativeStepDirection
I don't like the name very much. It is long and we don't use "step direction" anywhere else. What about changing it to one of
```go
ErrNonDescentDirection
ErrNonDecreasingDirection
ErrInvalidDirection // Mmm, not very descriptive
```
?

vladimir-ch:
Closes #124, I went with `ErrNonDescentDirection`, "descent direction" seems more common than "decrease direction".

I also renamed `ErrLinesearchFailure` to `ErrLinesearcherFailure`, slightly updated some error comments and reordered them. 

In an effort towards stabilization of the API we should review the error variables. For example, ErrLinesearcherFailure seems too broad and not very descriptive.

PTAL @btracey

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -10,28 +10,29 @@ var (
 	// ErrInf signifies the initial function value is Inf.
 	ErrInf = errors.New("optimize: initial function value is Inf")
 
-	// ErrGradInf signifies the initial function value is Inf.
-	ErrGradInf = errors.New("optimize: initial gradient is Inf")
-
-	// ErrLinesearchFailure signifies the linesearch has iterated too many
-	// times. This may occur if the gradient tolerance is set too low.
-	ErrLinesearchFailure = errors.New("linesearch: failed to converge")
-
 	// ErrNaN signifies the initial function value is NaN.
 	ErrNaN = errors.New("optimize: initial function value is NaN")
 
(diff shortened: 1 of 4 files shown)


errors.go
```diff
@@ -10,28 +10,29 @@ var (
...
-	// times. This may occur if the gradient tolerance is set too low.
-	ErrLinesearchFailure = errors.New("linesearch: failed to converge")
-
 	// ErrNaN signifies the initial function value is NaN.
 	ErrNaN = errors.New("optimize: initial function value is NaN")
 
-	// ErrGradNaN signifies the initial function value is NaN.
-	ErrGradNaN = errors.New("optimize: initial gradient is NaN")
+	// ErrGradInf signifies that a component of the initial gradient is Inf.
+	ErrGradInf = errors.New("optimize: initial gradient is Inf")
```
btracey: @kortschak, what's the style here? "Inf" "inf" or "∞"?

btracey:
LGTM

errors.go
kortschak: From looking through the stanard library, it looks like "optimize: initial gradient is infinite". The general rule is "Use English".

PR #134, opened 2015-09-25
Remove Linesearcher.Finished()

vladimir-ch:
* Removes `Linesearcher.Finished()`
* Changes the order of return values from `Linesearcher.Iterate()`. Operation is the main value returned, so put it first.
* Slightly cleans up `Backtracking` and `Bisection`
* And tries to update the docs for `Linesearcher` ...

@btracey Please take a look

### DIFF
--- a/backtracking.go
+++ b/backtracking.go
@@ -31,6 +31,8 @@ type Backtracking struct {
 	stepSize float64
 	initF    float64
 	initG    float64
+
+	lastOp Operation
 }
 
 func (b *Backtracking) Init(f, g float64, step float64) Operation {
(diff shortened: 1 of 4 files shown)


backtracking.go
```diff
@@ -57,17 +59,27 @@ func (b *Backtracking) Init(f, g float64, step float64) Operation {
...
 	b.stepSize *= b.Decrease
 	if b.stepSize < minimumBacktrackingStepSize {
-		return 0, NoOperation, ErrLinesearchFailure
+		return b.next(NoOperation, ErrLinesearchFailure)
 	}
-	return b.stepSize, FuncEvaluation, nil
+	return b.next(FuncEvaluation, nil)
+}
+
+func (b *Backtracking) next(op Operation, err error) (Operation, float64, error) {
```
btracey: This function is simple and only happens twice. I think it's better to just put this in line.

interfaces.go
```diff
@@ -65,24 +65,35 @@ type Statuser interface {
 // (approximate) minimum of the objective function along the search direction
 // dir_k starting at the most recent location x_k, i.e., it tries to minimize
 // the function
+//
```
btracey: Don't need these newlines.

btracey:
Nice. LGTM.

backtracking.go
vladimir-ch: Three times, but nevermind, inlined.

interfaces.go
vladimir-ch: I like the formulas to have some margin around them, similarly to mathematical formulas in papers. It looks better to me and the formula stands out, which is often good, for example in this case we refer to it at several places below. But ok, I removed the newlines.

PR #135, opened 2015-09-26
Swap arguments to Grad and Hess

vladimir-ch:
@btracey PTAL

### DIFF
--- a/types.go
+++ b/types.go
@@ -162,9 +162,9 @@ func (p Problem) satisfies(method Method) error {
 // If Recorder is nil, no information will be recorded.
 type Settings struct {
 	UseInitialData  bool            // Use supplied information about the conditions at the initial x.
-	InitialValue    float64         // Func(x) at the initial x.
-	InitialGradient []float64       // Grad(x) at the initial x.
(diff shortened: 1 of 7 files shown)


types.go
```diff
@@ -127,11 +127,11 @@ type Problem struct {
 
 	// Grad evaluates the gradient at x and stores the result in-place in grad.
 	// Grad must not modify x.
-	Grad func(x []float64, grad []float64)
+	Grad func(grad []float64, x []float64)
 
 	// Hess evaluates the Hessian at x and stores the result in-place in hess.
 	// Hess must not modify x.
-	Hess func(x []float64, hess *mat64.SymDense)
+	Hess func(hess *mat64.SymDense, x []float64)
```
btracey: Can we change this to MutableSym while we're breaking everything?

btracey:
LGTM ignoring the interface change.

types.go
vladimir-ch: AFAICT we don't have MutableSym (yet).

btracey: https://github.com/gonum/matrix/pull/242

vladimir-ch: Argh ... and @kortschak complains that _he_ says foolish things ...

PR #136, opened 2015-09-27
Replace Inf with infinite in errors

vladimir-ch:
@kortschak @btracey

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -7,14 +7,14 @@ package optimize
 import "errors"
 
 var (
-	// ErrInf signifies the initial function value is Inf.
-	ErrInf = errors.New("optimize: initial function value is Inf")
+	// ErrInf signifies the initial function value is infinite.
+	ErrInf = errors.New("optimize: initial function value is infinite")
 
 	// ErrNaN signifies the initial function value is NaN.
 	ErrNaN = errors.New("optimize: initial function value is NaN")
 
-	// ErrGradInf signifies that a component of the initial gradient is Inf.
-	ErrGradInf = errors.New("optimize: initial gradient is Inf")
+	// ErrGradInf signifies that a component of the initial gradient is infinite.
+	ErrGradInf = errors.New("optimize: initial gradient is infinite")
 
 	// ErrGradNaN signifies that a component of the initial gradient is NaN.
 	ErrGradNaN = errors.New("optimize: initial gradient is NaN")


btracey:
lgtm

kortschak:
The approach used in math/big.Float with ErrNaN is quite nice. I'm just looking to see if that fits here.

kortschak:
... or our Condition error in mat64.

kortschak:
Strawman counter - not sure whether the float should be visible, it allows things like (otherwise requiring an additional method to return the invalid value):

```
switch err := err.(type) {
case ErrFunc:
	if math.IsNaN(float64(err)) {
		panic("terrible error")
	}
	return fmt.Errorf("bad error: %v", err)
case ErrGrad:
	if math.IsNaN(float64(err)) {
		panic("terrible error")
	}
	return fmt.Errorf("bad error: %v", err)
}
```

```
diff --git a/errors.go b/errors.go
index 94c503b..0ea77af 100644
--- a/errors.go
+++ b/errors.go
@@ -4,21 +4,12 @@
 
 package optimize
 
-import "errors"
+import (
+	"errors"
+	"math"
+)
 
 var (
-	// ErrInf signifies the initial function value is Inf.
-	ErrInf = errors.New("optimize: initial function value is Inf")
-
-	// ErrNaN signifies the initial function value is NaN.
-	ErrNaN = errors.New("optimize: initial function value is NaN")
-
-	// ErrGradInf signifies that a component of the initial gradient is Inf.
-	ErrGradInf = errors.New("optimize: initial gradient is Inf")
-
-	// ErrGradNaN signifies that a component of the initial gradient is NaN.
-	ErrGradNaN = errors.New("optimize: initial gradient is NaN")
-
 	// ErrZeroDimensional signifies an optimization was called with an input of length 0.
 	ErrZeroDimensional = errors.New("optimize: zero dimensional input")
 
@@ -36,3 +27,33 @@ var (
 )
+
+// ErrFunc is returned when an invalid initial function value is invalid. The error
+// state may be either +Inf or NaN. ErrFunc satisfies the error interface.
+type ErrFunc float64
+
+func (err ErrFunc) Error() string {
+	switch {
+	case math.IsInf(float64(err), 1):
+		return "optimize: initial function value is infinite"
+	case math.IsNaN(float64(err)):
+		return "optimize: initial function value is NaN"
+	default:
+		panic("optimize: bad ErrFunc")
+	}
+}
+
+// ErrFunc is returned when an invalid initial gradient is invalid. The error state
+// may be either ±Inf or NaN. ErrGrad satisfies the error interface.
+type ErrGrad float64
+
+func (err ErrGrad) Error() string {
+	switch {
+	case math.IsInf(float64(err), 0):
+		return "optimize: initial gradient is infinite"
+	case math.IsNaN(float64(err)):
+		return "optimize: initial gradient is NaN"
+	default:
+		panic("optimize: bad ErrGrad")
+	}
+}
diff --git a/local.go b/local.go
index 962f2ee..188458e 100644
--- a/local.go
+++ b/local.go
@@ -290,18 +290,12 @@ func getStartingLocation(p *Problem, method Method, initX []float64, stats *Stat
 		evaluate(p, loc, eval, stats)
 	}
 
-	if math.IsNaN(loc.F) {
-		return loc, ErrNaN
-	}
-	if math.IsInf(loc.F, 1) {
-		return loc, ErrInf
+	if math.IsNaN(loc.F) || math.IsInf(loc.F, 1) {
+		return loc, ErrFunc(loc.F)
 	}
 	for _, v := range loc.Gradient {
-		if math.IsInf(v, 0) {
-			return loc, ErrGradInf
-		}
-		if math.IsNaN(v) {
-			return loc, ErrGradNaN
+		if math.IsInf(v, 0) || math.IsNaN(v) {
+			return loc, ErrGrad(v)
 		}
 	}
 ```

vladimir-ch:
I think it is nice. The amount of new code in errors.go seems like a lot but 1) The checking code in getStartingLocation gets simpler 2) We have to do this check to be sure that we have valid initial data. I would rename the types to ErrInit{Func,Grad}, it seems more descriptive and accurate. In the Grad error it would be perhaps useful to capture also the component where the invalid value occured. @btracey ?

Also, what does "strawman counter" mean?

kortschak:
Strawman is something to put up that you expect might get knocked down as a starting point for discussion. Counter - short for counter proposal.

kortschak:
BTW Is it true that loc.F being -Inf is OK? The check for math.IsInf only looks for positive.

vladimir-ch:
It is ok in the sense that we cannot go lower than that, so the minimization should not even start. It is true that we probably do not handle this case (rather unlikely case IMO) well in Local().

vladimir-ch:
Strawman counter - with the explanation it seems embarassingly obvious.

kortschak:
Do you want me to send the change above or will you do it, or is it not wanted?

vladimir-ch:
Yes, please submit it with the suggested changes. Thanks.

Issue #127, opened 2015-09-24
Linesearchers should check function value before asking for gradient


btracey:
Many problems have a gradient that's expensive to compute. The line searchers we implement should check that the function value has decreased before asking for the gradient.

vladimir-ch:
Backtracking returns only FuncEvaluation, but at present Bisection needs both at the same time. Splitting it will require some changes and perhaps a bit of thought. If the step satisfies Armijo, then it is easy, just return GradEvaluation. If not, then it is not so clear. If maxStep has not been found yet, then probably the right thing is to halve the step and try again with another FuncEvaluation. If the step has already been bounded, then ... ?

Anyway, when this is done, I will be _this_ close from suggesting to drop the ORability of evaluation operations. Just saying (although I am afraid it was me who came up with the idea).

vladimir-ch:
I tried to come up with something for Bisection that would actually save some work if a step does not satisfy the Armijo condition. But needing both values simultaneously is essential for the algorithm, so either we don't save work or it will be a different algorithm. Do you see another way, @btracey ?

If we just want to issue FuncEvaluation and GradEvalaution separately, it can be done easily, though.

btracey:
That's exactly what I have in mind. Request function value. If it passes, request gradient. If not, update

vladimir-ch:
But update needs gradient too, so we don't save anything.

vladimir-ch:
Continuing the previous discussion from closed #137:

> I do think that our line searches should evaluate individually (for example, so gradient computations can be saved).

... One was to see if gradient computations can be saved if we know that the point does not satisfy even Armijo. It is obvious to me now that they cannot, the algorithm simply needs the derivative at the midpoint to decide which side to subdivide. We could modify the algorithm but then it would not be the same thing, it is not clear how to modify it and whether it is worth it. If we modified the bracketing phase not to use g, only f, then we could save them during that phase because the action does not need g, it is always: go forward.

... but Backtracking does not ask for the gradient and Bisection cannot do without it.

> That's not true on Bisection. If the function value is higher, then we always know which direction to push. If the function value is lower then the gradient is needed to decide which way to slide.

The very first check in Bisection is g < 0 so as it is now we always need the derivative. We can remove it, though, and make the bracketing phase derivative-free (slide forward until the midpoint has lower value than the two end points). Perhaps we should. Also, shouldn't we fail earlier than when b.step == step? Just asking.

btracey:
[...]

PR #137, opened 2015-09-27
Separate evaluations for Bisection

vladimir-ch:
No work is saved, but with this our Linesearchers don't need oring of Evaluations. Then the last place using it would be LinesearchMethod. @btracey PTAL

### DIFF
--- a/bisection.go
+++ b/bisection.go
@@ -13,18 +13,19 @@ import "math"
 type Bisection struct {
 	GradConst float64
 
-	minStep  float64
-	maxStep  float64
-	currStep float64
-
 	initF float64
-	minF  float64
-	maxF  float64
-	f     float64
+	initG float64
 
-	initGrad float64
-	minGrad  float64
-	maxGrad  float64
+	minStep float64
+	minF    float64
+	minG    float64
+
+	maxStep float64
+	maxF    float64
+	maxG    float64
+
+	step float64
+	f    float64
 
 	lastOp Operation
 }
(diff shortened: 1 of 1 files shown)


btracey:
What is the goal of this change? It doesn't seem to me like ORing of evaluations is a cost. The reverse communication interface can be tricky to reason about as a Method implementer, and ORing makes that process simpler. Do you disagree?

btracey:
I do think that our line searches should evaluate individually (for example, so gradient computations can be saved). Is this a step towards resolving that problem, or in changing Optimize behavior?

vladimir-ch:
There were two motivations for me to start this.

One was to see if gradient computations can be saved if we know that the point does not satisfy even Armijo. It is obvious to me now that they cannot, the algorithm simply needs the derivative at the midpoint to decide which side to subdivide. We could modify the algorithm but then it would not be the same thing, it is not clear how to modify it and whether it is worth it. If we modified the bracketing phase not to use g, only f, then we could save them during that phase because the action does not need g, it is always: go forward.

The second reason was that allowing ORing of evaluating operation slightly complicates the code (String(), checking of returned operations) and documenting it, not that it is a computational cost. So I tried to see how it would work out without ORing. In Bisection, the result is this. Yes, ORing is convenient, but I thought that if someone grasps the reverse communication well enough to write a Method, then issuing separate evaluations is not a problem.

It is really just an experiment and a base for (this) discussion. Moreover, you filed #127, but Backtracking does not ask for the gradient and Bisection cannot do without it. So if we reject this, then we can close #127.

btracey:
That's not true on Bisection. If the function value is higher, then we always know which direction to push. If the function value is lower then the gradient is needed to decide which way to slide. 

Are you sure that the final result is net less complexity by removing the Oring? You're right that the string method is complicated, and checking that it's an evaluation is somewhat complicated. On the other hand, your code in Linesearch for filling Location with a valid location is very nice (complementEval). Removing Orable evaluations makes that more complicated.

btracey:
In particular, for example, your code for the "evaluate" function is extremely clean, even with the Orable complexity. Finding the words to document the precise behavior is tough, but I think conceptually it's clear what you can and cannot combine. I'm also not convinced String is that much easier since we want to help debug anyway.

If you think the orable is that complicated, we can get rid of it. It just seems to me like you found a design that removes all of the unnecessary complexity, and we are left with a very clean expression mechanism.

vladimir-ch:
> That's not true on Bisection. If the function value is higher, then we always know which direction to push. If the function value is lower then the gradient is needed to decide which way to slide.

The very first check in Bisection is g < 0 so as it is now we always need the derivative. We can remove it, though, and make the bracketing phase derivative-free. Perhaps we should. Also, shouldn't we fail earlier than when b.step == step? Just asking.

> On the other hand, your code in Linesearch for filling Location with a valid location is very nice (complementEval). Removing Orable evaluations makes that more complicated.

That's true. This and your other arguments convince me that ORability should stay. I will close this PR and resubmit at least the tiny field names clean up separately.

Let's continue the discussion about Bisection in #127.

PR #138, opened 2015-09-28
optimize: make initialization errors hold their cause

kortschak:
Changes discussed in https://github.com/gonum/optimize/pull/136#issuecomment-143515739

@vladimir-ch @btracey Please take a look.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -4,21 +4,13 @@
 
 package optimize
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+	"math"
+)
 
 var (
-	// ErrInf signifies the initial function value is infinite.
-	ErrInf = errors.New("optimize: initial function value is infinite")
-
-	// ErrNaN signifies the initial function value is NaN.
-	ErrNaN = errors.New("optimize: initial function value is NaN")
-
-	// ErrGradInf signifies that a component of the initial gradient is infinite.
-	ErrGradInf = errors.New("optimize: initial gradient is infinite")
-
-	// ErrGradNaN signifies that a component of the initial gradient is NaN.
-	ErrGradNaN = errors.New("optimize: initial gradient is NaN")
-
 	// ErrZeroDimensional signifies an optimization was called with an input of length 0.
 	ErrZeroDimensional = errors.New("optimize: zero dimensional input")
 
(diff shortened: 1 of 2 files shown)


errors.go
```diff
@@ -36,3 +27,33 @@ var (
 	// due to floating-point arithmetic.
 	ErrNoProgress = errors.New("linesearch: no change in location after Linesearcher step")
 )
+
+// ErrFuncInit is returned when an initial function value is invalid. The error
+// state may be either +Inf or NaN. ErrFuncInit satisfies the error interface.
+type ErrFuncInit float64
```
vladimir-ch: Which is better, `ErrFuncInit` or `ErrInitFunc`? The latter sounds better to me.

errors.go
```diff
@@ -36,3 +27,33 @@ var (
...
+	case math.IsNaN(float64(err)):
+		return "optimize: initial function value is NaN"
+	default:
+		panic("optimize: bad ErrFuncInit")
+	}
+}
+
+// ErrGradInit is returned when an initial gradient is invalid. The error state
+// may be either ±Inf or NaN. ErrGradInit satisfies the error interface.
+type ErrGradInit float64
```
vladimir-ch: Would it be overkill to make the type
```go
type ErrGradInit struct {
    v     float64 // Invalid value
    index int     // Index of the value
}
```
(probably with the fields exported)? Seems like the index could be useful information.

errors.go
kortschak: I agree. I actually like it without at all, but that's from a perspective of someone who doesn't yet use this package.

errors.go
kortschak: SGTM

btracey: Alternatively, we could just have one ErrInit type. It could have the value and the problem subset (function, gradient, etc.) as a field.

vladimir-ch: Re. having Init in the name: I don't feel strongly either way. ErrFunc, ErrGrad are nice and short, but they suggest that they can be used for any invalid values returned from the objective function  at any point while (at least at the moment) they are specific to initial data. For example Problem.Status might want to return them. Of course, in principle ErrFunc etc can do it but I think it should not be necessary. All this optimization stuff we have at the moment assumes well behaved functions, so that if we start from valid initial data, there always is a sufficiently small neighborhood where the function also returns valid data. If not, we can backtrack closer to previous location and retry until we eventually make no progress and then ErrNoProgress. We do not do it yet, but we should. Fixing Backtracking to do it is trivial, Bisection, mmmm, not sure off the top of my head. Anyway, I'm deviating from the topic.

Re. one ErrInit type: Nice idea. One type, one Error method. Perhaps I am slightly leaning towards this side? We could abuse Operation to hold where the invalid value comes from but it seems that the fields should not be exported in this case.

Also, now we are checking only the value and the gradient. Should we check also the Hessian? Consistency says so, but what if one day we have Newton for "large" problems? Also, do you plan to check also the constraints?

kortschak:
>ErrFunc, ErrGrad are nice and short, but they suggest that they can be used for any invalid values returned from the objective function at any point while (at least at the moment) they are specific to initial data.

The documentation makes this reasonably clear and if the API does need to change in this regard, leaving the name open makes that change non-breaking.

>One type, one Error method.

That essentially take the situation to what is in current master - there is one type with a variety of values.

btracey:
I'm thinking now that the whole approach to error handling may be misguided. 

- I'm thinking that Local is the wrong place to do initial value handling. Instead, this should be handled by specific Methods. I would imagine that there are many Methods which can handle a starting place of inf (Nelder Mead may already be able to do it).

- Either way, I do think there is utility in explicitly declaring that the error occurred in initialization. An initialization error is purely the fault of the user. The specifed problem does not work with the recommended algorithm. An error later in the optimization is either indicative of a problem with the code of the Method, the code of the function (if the gradients are wrong), or some interaction with the method and the function (say the function exhibits numerical noise and so cannot converge as far as requested).

vladimir-ch:
> I'm thinking that Local is the wrong place to do initial value handling. Instead, this should be handled by specific Methods.

Method.Init() returns an error, so we can move the check there if we want.

> An initialization error is purely the fault of the user. The specifed problem does not work with the recommended algorithm.

Yes, but the reason might be slightly different: in spite of assumptions the function is not defined everywhere and the user just provided a bad starting point. It is not so difficult to make it happen. For example, functions.HelicalValley panics for points on the plane x = 0 (it should probably return a NaN instead) and the zero vector is such a tempting starting location. Still, it is the fault of the user.

Anyway, we can postpone this discussion and continue it later. I like what is in this PR now, so LGTM. Thanks, @kortschak and sorry for the detour.

btracey:
Sgtm. I'll submit a follow-up

PR #142, opened 2015-10-01
Improve Bisection to command GradEvaluation only when necessary

vladimir-ch:
NOT (yet) FOR SUBMISSION

This addresses #127: Request GradEvaluation only if the step satisfies Armijo condition or if we have found a new lowest value.

Several tests fail with ErrLinesearcherFailure, each of them with tolerances between 1e-12 ~ 1e-9 and very close to the minimum. One test (VariablyDimensioned with dim = 10000) fails with ErrNoProgress at the beginning of an iteration (with the very first trial step), but also very close to the minimum (but this one has grad tol 1e-5). Either loosen the tolerances or improve the stopping criterion for Bisection. Could you please take a look, @btracey ?

### DIFF
--- a/bisection.go
+++ b/bisection.go
@@ -22,8 +22,6 @@ type Bisection struct {
 	maxF  float64
 
 	initGrad float64
-	minGrad  float64
-	maxGrad  float64
 
 	lastOp Operation
 }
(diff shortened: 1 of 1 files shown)


btracey:
I hope you don't mind, but I just created #143 . It is merely a re-working of the current Bisection logic, and passes all of the tests.

vladimir-ch:
Not at all. I think the logic is the same and if I changed how the steps are expanded during bracketing, the same evaluations would be requested. I will comment further in #143.

PR #143, opened 2015-10-01
Modify bisection so it only requests the gradient when appropriate

btracey:


btracey:
Fixes #127

bisection.go
```diff
@@ -74,63 +67,55 @@ func (b *Bisection) Iterate(f, g float64) (Operation, float64, error) {
...
+		b.lastOp = MajorIteration
+		return b.lastOp, b.currStep, nil
+	}
+	if math.IsInf(b.maxStep, 1) {
+		// The function value is lower. If the gradient is positive, an upper bound
+		// of the minimum been found. If the gradient is negative, search farther
+		// in that direction.
+		if g > 0 {
+			b.maxStep = b.currStep
+			b.maxF = f
```
vladimir-ch: We use f even though the last operation was GradEvaluation. We should either document that Linesearchers must receive the previous function value after requesting GradEvaluation or (probably better) we should store f after FuncEvaluation and then we don't have to assume anything.

bisection.go
```diff
@@ -74,63 +67,55 @@ func (b *Bisection) Iterate(f, g float64) (Operation, float64, error) {
...
 		} else {
 			b.minStep = b.currStep
 			b.minF = f
-			b.minGrad = g
 		}
+		return b.nextStep((b.minStep + b.maxStep) / 2)
+	}
+	// The function value was lower. Check if this location is sufficient to
+	// converge the linesearch, otherwise iterate.
+	if StrongWolfeConditionsMet(f, g, minF, b.initGrad, b.currStep, 0, b.GradConst) {
```
vladimir-ch: Wouldn't it be better replace this with an explicit check? Only the curvature condition needs to be checked and we wouldn't have to repeatedly call Abs on the initial derivative (not that it is a big cost). We will have to change the detection of convergence anyway to deal with the floating point noise.

bisection.go
```diff
@@ -74,63 +67,55 @@ func (b *Bisection) Iterate(f, g float64) (Operation, float64, error) {
...
+		// of the minimum been found. If the gradient is negative, search farther
+		// in that direction.
+		if g > 0 {
+			b.maxStep = b.currStep
+			b.maxF = f
+			return b.nextStep((b.minStep + b.maxStep) / 2)
+		}
+		b.minStep = b.currStep
+		b.minF = f
+		return b.nextStep(b.currStep * 2)
```
vladimir-ch: It seems to me that your version passing the tests is just a coincidence. Use `2+1e-12` and two tests will fail. Use `2+1e-6` and three tests will fail, only one will be common, two will be different. It suggests that we really have to take care of the floating point.

bisection.go
```diff
@@ -146,6 +131,7 @@ func (b *Bisection) nextStep(step float64) (Operation, float64, error) {
 		return b.lastOp, b.currStep, ErrLinesearcherFailure
 	}
 	b.currStep = step
-	b.lastOp = FuncEvaluation | GradEvaluation
+	//b.lastOp = FuncEvaluation | GradEvaluation
```
vladimir-ch: Remove this line

bisection.go
btracey: Yes, it would. The floating point issue was going to be my next PR. I was trying to have this change match the original code.

In either event, I would like to keep the StrongWolfe function itself. It won't change, but it's illustrative of what the conditions are.

bisection.go
btracey: I agree with your conclusion. It's not so much a coincidence as we've tuned the test tolerances so they match. The "lack of robustness" is due to floating point noise I'm pretty sure.

bisection.go
btracey: Fixed.

bisection.go
vladimir-ch: Ok

bisection.go
vladimir-ch: Agreed

vladimir-ch:
LGTM once the commented line is removed

vladimir-ch:
Btw, I like how our recent changes gradually reduce the runtime of the tests. This change helps especially CG, its search directions are badly scaled and the linesearch takes a couple of iterations (7-10 on average last time I checked, of course depending on the problem).

PR #144, opened 2015-10-02
Make function decrease check based on a tolerance

btracey:


btracey:
This is the code that needs to change, but I'm not sure about the comment. The current wording seems overly complicated. I'm not sure backtracking should have it. The whole point in backtracking is to find a sufficient function decrease.

bisection.go
```diff
@@ -4,15 +4,30 @@
...
+	"github.com/gonum/floats"
+)
 
 // Bisection is a Linesearcher that uses a bisection to find a point that
 // satisfies the strong Wolfe conditions with the given gradient constant and
 // function constant of zero. If GradConst is zero, it will be set to a reasonable
 // value. Bisection will panic if GradConst is not between zero and one.
+// EqTol is used to smooth out floating point noise when testing the Wolfe
+// conditions. The Wolfe condition will be considered met if the function value
+// is less than or within a tolerance of the starting value of the line
```
vladimir-ch: This is not true, we compare to the value on the left side of the interval (Is it even correct? The original code compared to the lowest value found).

There is not a suitable line to put the comment on, but `b.initF` field and `minF` variable are not used in this version, is it on purpose?

bisection.go
```diff
@@ -4,15 +4,30 @@
...
+// search. If EqTol is zero, it will be defaulted to 1e-10, and if it is negative
+// a strict less than or equal to comparison will be used. A larger value of eqTol
```
vladimir-ch: We will have to get the stopping criterion right first and then review the comment, but `strict` here seems superfluous.

bisection.go
```diff
@@ -72,7 +92,7 @@ func (b *Bisection) Iterate(f, g float64) (Operation, float64, error) {
 		// See if the function value is good enough to make progress. If it is,
 		// evaluate the gradient. If not, set it to the upper bound if the bound
 		// has not yet been found, otherwise iterate toward the minimum location.
-		if f <= b.minF {
+		if f <= b.minF || floats.EqualWithinAbsOrRel(f, b.minF, b.eqTol, b.eqTol) {
```
vladimir-ch: I wonder if the length of the interval shouldn't play a role here.

bisection.go
btracey: Nope, should be minF.

vladimir-ch:
The test
```
F: VariablyDimensioned
Dim: 10000
GradientThreshold: 1e-05
```
fails at random with LBFGS as the expansion factor is varied. It fails with `ErrNoProgress` on the first linesearcher step, which means that step 1 is accepted but x + grad == x. Is it fixable or is 1e-5 too stringent for this functions?

bisection.go
btracey: I don't think so. The comparison to minF is to keep it in the lowest base found .The floating point noise could crop up there.

vladimir-ch: I mean that we don't care about floating point much until the interval is short.

bisection.go
btracey: Maybe we should be vague in the comment. It's just a parameter to protect against numerical noise at the cost of missing a minima.

bisection.go
```diff
@@ -4,25 +4,38 @@
...
+
+	"github.com/gonum/floats"
+)
 
 // Bisection is a Linesearcher that uses a bisection to find a point that
 // satisfies the strong Wolfe conditions with the given gradient constant and
 // function constant of zero. If GradConst is zero, it will be set to a reasonable
 // value. Bisection will panic if GradConst is not between zero and one.
+// EqTol is used to smooth out floating point noise when testing the Wolfe
+// conditions. The Wolfe condition will be considered met if the function value
```
vladimir-ch: I would replace Wolfe with sufficient decrease or Armijo in these two lines.

bisection.go
```diff
@@ -4,25 +4,38 @@
...
+// is less than or within a tolerance of the starting value of the line
+// search. If EqTol is zero, it will be defaulted to 1e-10, and if it is negative
+// a strict less than or equal to comparison will be used. A larger value of eqTol
+// will be more robust to numeric fluctuations but will miss real purturbations
+// in the objective function while a smaller value of eqTol will be more sensative
```
vladimir-ch: If you keep this, then `sensative -> sensitive`.

When you edit this, could you move the whole EqTo to the field in the struct? And while you are at it, do the same for GradConst?

vladimir-ch:

@btracey , should we port this to the new repository or has it been done in the meantime?

PR #145, opened 2015-10-05
Pass a copy of Location.X to evaluating routines

### LINKED ISSUE #129: Pass a copy of X to evaluation routines
Pass a copy of X to evaluation routines so that `Problem` cannot overwrite it either by mistake or on (misguided) purpose.

vladimir-ch:
@btracey PTAL

Fixes #129


btracey:
LGTM

PR #147, opened 2015-11-06
Use mat64 instead of floats in BFGS

vladimir-ch:
I think that with mat64 the code is cleaner and clearer.

Also, don't update the inverse Hessian if `sDotY` is zero (to avoid division by zero).

@btracey @kortschak PTAL

### DIFF
--- a/bfgs.go
+++ b/bfgs.go
@@ -5,7 +5,8 @@
 package optimize
 
 import (
-	"github.com/gonum/floats"
+	"math"
+
 	"github.com/gonum/matrix/mat64"
 )
 
@@ -15,28 +16,23 @@ import (
 // super-linear convergence when in proximity to a local minimum. It has memory
 // cost that is O(n^2) relative to the input dimension.
 type BFGS struct {
-	// Linesearcher is used for selecting suitable steps along the descent
-	// direction. Accepted steps should satisfy the strong Wolfe conditions.
+	// Linesearcher selects suitable steps along the descent direction.
+	// Accepted steps should satisfy the strong Wolfe conditions.
 	// If Linesearcher == nil, an appropriate default is chosen.
 	Linesearcher Linesearcher
 
 	ls *LinesearchMethod
 
-	x    []float64 // location of the last major iteration
-	grad []float64 // gradient at the last major iteration
 	dim  int
-
-	// Temporary memory
-	y      []float64
-	yVec   *mat64.Vector
-	s      []float64
-	sVec   *mat64.Vector
-	tmp    []float64
-	tmpVec *mat64.Vector
+	x    mat64.Vector // Location of the last major iteration.
+	grad mat64.Vector // Gradient at the last major iteration.
+	s    mat64.Vector // Difference between locations in this and the previous iteration.
+	y    mat64.Vector // Difference between gradients in this and the previous iteration.
+	tmp  mat64.Vector
 
 	invHess *mat64.SymDense
 
-	first bool // Is it the first iteration (used to set the scale of the initial hessian)
+	first bool // Indicator of the first iteration.
 }
 
 func (b *BFGS) Init(loc *Location) (Operation, error) {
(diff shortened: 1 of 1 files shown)


bfgs.go
```diff
@@ -5,38 +5,34 @@
...
-// an estimate of the inverse-Hessian of the function. It exhibits super-linear
-// convergence when in proximity to a local minimum. It has memory cost that is
-// O(n^2) relative to the input dimension.
+// BFGS implements the Broyden–Fletcher–Goldfarb–Shanno optimization method. It
+// is a quasi-Newton method that performs successive rank-one updates to an
+// estimate of the inverse Hessian of the objective function. It exhibits
+// super-linear convergence when in proximity to a local minimum. It has memory
+// cost that is O(n^2) relative to the input dimension.
 type BFGS struct {
+	// Linesearcher is used for selecting suitable steps along the descent
```
btracey: Linesearcher selects suitable steps

bfgs.go
```diff
@@ -5,38 +5,34 @@
...
-	// Temporary memory
-	y      []float64
-	yVec   *mat64.Vector
-	s      []float64
-	sVec   *mat64.Vector
-	tmp    []float64
-	tmpVec *mat64.Vector
+	x    mat64.Vector // Location of the last major iteration.
+	grad mat64.Vector // Gradient at the last major iteration.
+	s    mat64.Vector // Difference between locations in the last two iterations.
```
btracey: In this and the previous iteration?

bfgs.go
```diff
@@ -59,65 +55,62 @@ func (b *BFGS) Iterate(loc *Location) (Operation, error) {
...
-	// Compute the step difference
 	// s = x_{k+1} - x_{k}
-	floats.SubTo(b.s, loc.X, b.x)
+	b.s.SubVec(x, &b.x)
+	// y = g_{k+1} - g_{k}
+	b.y.SubVec(grad, &b.grad)
 
-	sDotY := floats.Dot(b.s, b.y)
-	sDotYSquared := sDotY * sDotY
+	sDotY := mat64.Dot(&b.s, &b.y)
```
btracey: Could you please open up an issue and/or PR to fast path Vector in Dot? We'd like to keep adding to the efficiency of mat64, and us needing it is a particularly good thing to add.

bfgs.go
```diff
@@ -128,34 +121,31 @@ func (b *BFGS) NextDirection(loc *Location, dir []float64) (stepSize float64) {
...
-	copy(b.x, loc.X)
-	copy(b.grad, loc.Gradient)
-
-	// Compute the new search direction
-	d := mat64.NewVector(b.dim, dir)
-	g := mat64.NewVector(b.dim, loc.Gradient)
-
-	d.MulVec(b.invHess, g) // new direction stored in place
-	floats.Scale(-1, dir)
+	if math.Abs(sDotY) != 0 {
```
btracey: Should this actually be an equality, or should it be less than some tolerance?

bfgs.go
```diff
@@ -128,34 +121,31 @@ func (b *BFGS) NextDirection(loc *Location, dir []float64) (stepSize float64) {
...
+		//
+		//  B_{k+1}^-1 = B_k^-1
+		//             + (s_k^T y_k + y_k^T B_k^-1 y_k) / (s_k^T y_k)^2 * (s_k s_k^T)
+		//             - (B_k^-1 y_k s_k^T + s_k y_k^T B_k^-1) / (s_k^T y_k).
+		//
+		// Note that y_k^T B_k^-1 y_k is a scalar, and that the third term is a
+		// rank-two update where B_k^-1 y_k is one vector and s_k is the other.
+		yBy := mat64.Inner(&b.y, b.invHess, &b.y)
+		b.tmp.MulVec(b.invHess, &b.y)
+		scale := (sDotY + yBy) / sDotY / sDotY
```
btracey: Replace this with (1 + yBy/sDotY) / sDotY. Should be better scaled.

btracey:
LGTM with the minor comments.

bfgs.go
vladimir-ch: I just did without having read this comment. There is a small glitch, however.

kortschak: Collision.

vladimir-ch: Ooops.

kortschak: Yours is better.

bfgs.go
vladimir-ch: Good question. I have seen the test `if sDotY > 1e-14 { update }` used elsewhere. If I plug in the same check, several test fail. I decrease the constant, the number of failed tests also decreases, but some obviously take much longer to finish. The smallest I tried was 1e-18, then all tests pass.

Since that constant is kind of heuristical and arbitrary, for the moment I would leave the test as is. It fixes the problem I had and all tests quickly pass.

bfgs.go
vladimir-ch: Done

bfgs.go
vladimir-ch: Done

bfgs.go
vladimir-ch: Excellent and depressing suggestion at the same time: it makes that test pass, otherwise the tolerance would have to be changed from 1e-10 to 1e-8!

PR #148, opened 2015-11-10
Add MoreThuente Linesearcher

vladimir-ch:
I like the names DecreaseFactor and CurvatureFactor. I would vote for using them also in the other Linesearchers.

As I mentioned before, CG is the method that most benefits from this Linesearcher, so I set it as the default. Maybe we should remake the CG tests, there are already many combinations and the Linesearchers add an extra dimension.

PTAL @btracey

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -27,6 +27,10 @@ var (
 	// progress because there is no change in location after Linesearcher step
 	// due to floating-point arithmetic.
 	ErrNoProgress = errors.New("linesearch: no change in location after Linesearcher step")
+
+	// ErrLinesearcherBound signifies that a Linesearcher reached a step that
+	// lies out of allowed bounds.
+	ErrLinesearcherBound = errors.New("linesearch: step out of bounds")
 )
 
 // ErrFunc is returned when an initial function value is invalid. The error
--- a/functions/functions.go
+++ b/functions/functions.go
@@ -1733,3 +1733,46 @@ func (f Plassmann) Grad(grad, x []float64) {
 		grad[0] += 1
 	}
 }
+
+// YanaiOzawaKaneko is an univariate convex function where the values of Beta1
+// and Beta2 control the curvature around the minimum. Far away from the
+// minimum the function approximates an absolute value function. Near the
+// minimum, the function can either be sharply curved or flat, controlled by
+// the parameter values.
+//
+// References:
+//  - More, J.J., and Thuente, D.J.: Line Search Algorithms with Guaranteed Sufficient Decrease.
+//    ACM Transactions on Mathematical Software 20(3) (1994), 286–307, eq. (5.4)
+//  - Yanai, H., Ozawa, M., and Kaneko, S.: Interpolation methods in one dimensional
+//    optimization. Computing 27 (1981), 155–163
+type YanaiOzawaKaneko struct {
+	Beta1 float64
+	Beta2 float64
+}
+
+func (f YanaiOzawaKaneko) Func(x []float64) float64 {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	a := x[0]
+	b1 := f.Beta1
+	b2 := f.Beta2
+	g1 := math.Sqrt(1+b1*b1) - b1
+	g2 := math.Sqrt(1+b2*b2) - b2
+	return g1*math.Sqrt((a-1)*(a-1)+b2*b2) + g2*math.Sqrt(a*a+b1*b1)
+}
+
+func (f YanaiOzawaKaneko) Grad(grad, x []float64) {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+	a := x[0]
+	b1 := f.Beta1
+	b2 := f.Beta2
+	g1 := math.Sqrt(1+b1*b1) - b1
+	g2 := math.Sqrt(1+b2*b2) - b2
+	grad[0] = g1*(a-1)/math.Sqrt(b2*b2+(a-1)*(a-1)) + g2*a/math.Sqrt(b1*b1+a*a)
+}
--- a/functions/functions.go
+++ b/functions/functions.go
@@ -1678,3 +1678,58 @@ func (ConcaveLeft) Grad(grad, x []float64) {
 	}
 	grad[0] = math.Pow(x[0]+0.004, 3) * (5*x[0] - 7.98)
 }
+
+// Plassmann implements an univariate oscillatory function where the value of L
+// controls the number of oscillations. The value of Beta controls the size of
+// the derivative at zero and the size of the interval where the strong Wolfe
+// conditions can hold. For small values of Beta this function represents a
+// difficult test problem for linesearchers also because the information based
+// on the derivative is unreliable due to the oscillations.
+//
+// References:
+//  More, J.J., and Thuente, D.J.: Line Search Algorithms with Guaranteed Sufficient Decrease.
+//  ACM Transactions on Mathematical Software 20(3) (1994), 286–307, eq. (5.3)
+type Plassmann struct {
+	L    float64 // Number of oscillations for |x-1| ≥ Beta.
+	Beta float64 // Size of the derivative at zero, f'(0) = -Beta.
+}
+
+func (f Plassmann) Func(x []float64) float64 {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	a := x[0]
+	b := f.Beta
+	l := f.L
+	r := 2 * (1 - b) / l / math.Pi * math.Sin(l*math.Pi/2*a)
+	switch {
+	case a <= 1-b:
+		r += 1 - a
+	case 1-b < a && a <= 1+b:
+		r += 0.5 * ((a-1)*(a-1)/b + b)
+	default: // a > 1+b
+		r += a - 1
+	}
+	return r
+}
+
+func (f Plassmann) Grad(grad, x []float64) {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+	a := x[0]
+	b := f.Beta
+	l := f.L
+	grad[0] = (1 - b) * math.Cos(l*math.Pi/2*a)
+	switch {
+	case a <= 1-b:
+		grad[0] -= 1
+	case 1-b < a && a <= 1+b:
+		grad[0] += (a - 1) / b
+	default: // a > 1+b
+		grad[0] += 1
+	}
+}
--- a/functions/functions.go
+++ b/functions/functions.go
@@ -1627,3 +1627,29 @@ func (Wood) Minima() []Minimum {
 		},
 	}
 }
+
+// ConcaveRight implements an univariate function that is concave to the right
+// of the minimizer which is located at x=sqrt(2).
+//
+// References:
+//  More, J.J., and Thuente, D.J.: Line Search Algorithms with Guaranteed Sufficient Decrease.
+//  ACM Transactions on Mathematical Software 20(3) (1994), 286–307, eq. (5.1)
+type ConcaveRight struct{}
+
+func (ConcaveRight) Func(x []float64) float64 {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	return -x[0] / (x[0]*x[0] + 2)
+}
+
+func (ConcaveRight) Grad(grad, x []float64) {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+	xSqr := x[0] * x[0]
+	grad[0] = (xSqr - 2) / (xSqr + 2) / (xSqr + 2)
+}
--- a/functions/functions.go
+++ b/functions/functions.go
@@ -1653,3 +1653,28 @@ func (ConcaveRight) Grad(grad, x []float64) {
 	xSqr := x[0] * x[0]
 	grad[0] = (xSqr - 2) / (xSqr + 2) / (xSqr + 2)
 }
+
+// ConcaveLeft implements an univariate function that is concave to the left of
+// the minimizer which is located at x=399/250=1.596.
+//
+// References:
+//  More, J.J., and Thuente, D.J.: Line Search Algorithms with Guaranteed Sufficient Decrease.
+//  ACM Transactions on Mathematical Software 20(3) (1994), 286–307, eq. (5.2)
+type ConcaveLeft struct{}
+
+func (ConcaveLeft) Func(x []float64) float64 {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	return math.Pow(x[0]+0.004, 4) * (x[0] - 1.996)
+}
+
+func (ConcaveLeft) Grad(grad, x []float64) {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+	grad[0] = math.Pow(x[0]+0.004, 3) * (5*x[0] - 7.98)
+}
--- a/linesearcher_test.go
+++ b/linesearcher_test.go
@@ -1,0 +1,136 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package optimize
+
+import (
+	"fmt"
+	"math"
+	"reflect"
+	"testing"
+
+	"github.com/gonum/optimize/functions"
+)
+
+func TestMoreThuente(t *testing.T) {
+	d := 0.001
+	c := 0.001
+	ls := &MoreThuente{
+		DecreaseFactor:  d,
+		CurvatureFactor: c,
+	}
+	testLinesearcher(t, ls, d, c, true)
+}
+
+func TestBisection(t *testing.T) {
+	c := 0.1
+	ls := &Bisection{
+		GradConst: c,
+	}
+	testLinesearcher(t, ls, 0, c, true)
+}
+
+func TestBacktracking(t *testing.T) {
+	d := 0.001
+	ls := &Backtracking{
+		FuncConst: d,
+	}
+	testLinesearcher(t, ls, d, 0, false)
+}
+
+type funcGrader interface {
+	Func([]float64) float64
+	Grad([]float64, []float64)
+}
+
+type linesearcherTest struct {
+	name string
+	f    func(float64) float64
+	g    func(float64) float64
+}
+
+func newLinesearcherTest(name string, fg funcGrader) linesearcherTest {
+	grad := make([]float64, 1)
+	return linesearcherTest{
+		name: name,
+		f: func(x float64) float64 {
+			return fg.Func([]float64{x})
+		},
+		g: func(x float64) float64 {
+			fg.Grad(grad, []float64{x})
+			return grad[0]
+		},
+	}
+}
+
+func testLinesearcher(t *testing.T, ls Linesearcher, decrease, curvature float64, strongWolfe bool) {
+	for i, prob := range []linesearcherTest{
+		newLinesearcherTest("Concave-to-the-right function", functions.ConcaveRight{}),
+		newLinesearcherTest("Concave-to-the-left function", functions.ConcaveLeft{}),
+		newLinesearcherTest("Plassmann wiggly function (l=39, beta=0.01)", functions.Plassmann{39, 0.01}),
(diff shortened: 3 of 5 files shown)


linesearcher_test.go
```diff
@@ -0,0 +1,177 @@
...
+	testLinesearcher(ls, d, 0, false, t)
+}
+
+type lsProblem struct {
+	name string
+	f    func(float64) float64
+	g    func(float64) float64
+}
+
+// newPlassmannFunc returns a test function labeled as (5.3) in
```
btracey: Shouldn't these be in functions? Even if specialized I think we/someone will find it useful to have an easily accessible set of functions. We could provide a one-D wrapper here or something.

vladimir-ch: Indeed. I will move them there.

vladimir-ch: Done. PTAL @btracey

errors.go
```diff
@@ -27,6 +27,10 @@ var (
 	// progress because there is no change in location after Linesearcher step
 	// due to floating-point arithmetic.
 	ErrNoProgress = errors.New("linesearch: no change in location after Linesearcher step")
+
+	// ErrMaximumStep signifies that a Linesearcher reached a maximum allowed
+	// step that is not satisfactory.
+	ErrMaximumStep = errors.New("linesearch: reached maximum step")
```
btracey: How about this is "ErrLinesearchBound". Some linesearchers have a minimum step, and this may as well cover both?

functions/functions.go
```diff
@@ -1627,3 +1627,144 @@ func (Wood) Minima() []Minimum {
...
+//  More, J.J., and Thuente, D.J.: Line Search Algorithms with Guaranteed Sufficient
+//  Decrease. ACM Transactions on Mathematical Software 20(3) (1994), 286–307
+// The function is concave to the right of the minimizer that is located at x=sqrt(2).
+type ConcaveRight struct{}
+
+func (ConcaveRight) Func(x []float64) float64 {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+
```
btracey: remove newline

functions/functions.go
```diff
@@ -1627,3 +1627,144 @@ func (Wood) Minima() []Minimum {
...
+}
+
+func (ConcaveRight) Grad(grad, x []float64) {
+	if len(x) != 1 {
+		panic("dimension of the problem must be 1")
+	}
+	if len(x) != len(grad) {
+		panic("incorrect size of the gradient")
+	}
+
```
btracey: remove newline

functions/functions.go
```diff
@@ -1627,3 +1627,144 @@ func (Wood) Minima() []Minimum {
...
+	grad[0] = math.Pow(x[0]+0.004, 3) * (5*x[0] - 7.98)
+}
+
+// Plassmann implements the univariate function labeled as (5.3) in
+//  More, J.J., and Thuente, D.J.: Line Search Algorithms with Guaranteed Sufficient
+//  Decrease. ACM Transactions on Mathematical Software 20(3) (1994), 286–307
+type Plassmann struct {
```
btracey: Is there a one sentence why this is interesting?

functions/functions.go
```diff
@@ -1627,3 +1627,144 @@ func (Wood) Minima() []Minimum {
...
+		grad[0] += 1
+	}
+}
+
+// YanaiOzawaKaneko implements the univariate function labeled as (5.4) in
+//  More, J.J., and Thuente, D.J.: Line Search Algorithms with Guaranteed Sufficient
+//  Decrease. ACM Transactions on Mathematical Software 20(3) (1994), 286–307
+// See also
+//  Yanai, H., Ozawa, M., and Kaneko, S.: Interpolation methods in one dimensional
+//  optimization. Computing 27 (1981), 155–163
```
btracey: One sentence why this is interesting?

morethuente.go
```diff
@@ -0,0 +1,383 @@
...
+// MoreThuente is a Linesearcher that finds steps that satisfy both the
+// sufficient decrease and curvature conditions (the strong Wolfe conditions).
+//
+// References:
+//  - More, J.J. and D.J. Thuente: Line Search Algorithms with Guaranteed Sufficient
+//    Decrease. ACM Transactions on Mathematical Software 20(3) (1994), 286-307
+type MoreThuente struct {
+	// DecreaseFactor is the constant factor in the sufficient decrease
+	// (Armijo) condition.
+	// It must be from the interval [0, 1). The default value is 0.
```
btracey: be in the interval

morethuente.go
```diff
@@ -0,0 +1,383 @@
...
+	DecreaseFactor float64
+	// CurvatureFactor is the constant factor in the Wolfe conditions. Smaller
+	// values result in a more exact line search.
+	// It must be from the interval [0, 1). If it is zero, it will be defaulted
```
btracey: be in the interval

btracey: Maybe we should say "A set value must be in the interval (0, 1). If it is zero..."

morethuente.go
```diff
@@ -0,0 +1,383 @@
...
+	// to 0.9.
+	CurvatureFactor float64
+	// StepTolerance is the tolerance for an acceptable step. The line search
```
btracey: This is too vague. "StepTolerance sets the minimum acceptable width for the linesearch interval. If the relative interval length is less than this value, ErrLinesearchFailure is returned.

morethuente.go
```diff
@@ -0,0 +1,383 @@
...
+	// returns ErrLinesearcherFailure if the relative width of the bracketed
+	// interval is less than this value.
+	// It must be non-negative. If it is zero, it will be defaulted to 1e-10.
+	StepTolerance float64
+
+	// MinimumStep is the minimum step that the linesearcher will take.
```
btracey: Should it be the case that StepTolerance is a relative value while Minimu and Maximum are absolute values? Seems odd.

btracey: "Defaults to no minimum (a value of 0)"?

linesearcher_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+			op := ls.Init(f0, g0, initStep)
+
+			var (
+				err  error
+				k    int
+				f, g float64
+				step float64
+			)
+		loop:
```
btracey: This is a nice test (and not just because of how easy it is under the new invariants!). We may want to think about generalizing it to all the line searchers, i.e.  an "acceptableStep" switch or something.  This would help us make the the behavior similar across the linesearchers, especially in regard to floating point noise.

btracey: This is good for now though until we know how that design should be. Maybe just a "TODO"

errors.go
vladimir-ch: Changed to `ErrLinesearcherBound`. At the moment I will leave it as an ordinary error but I can imagine that storing the step and the bounds could be potentially useful.

functions/functions.go
vladimir-ch: Done

functions/functions.go
vladimir-ch: Done.

functions/functions.go
vladimir-ch: Done. This function is the most interesting of the four.

functions/functions.go
vladimir-ch: Done.

morethuente.go
vladimir-ch: Done.

morethuente.go
vladimir-ch: Done both.

morethuente.go
vladimir-ch: Thanks. Done.

morethuente.go
vladimir-ch: I think that Minimum and MaximumStep being absolute is not odd. They are a priori bounds on the step independent of the actual value of the step. Our current optimizers are unconstrained and do not bound the step. But for example Nocedal's LBFGS-B code uses the maximum bound to limit the step length in the given direction and that was also my motivation for including it here.

linesearcher_test.go
vladimir-ch: I don't understand these two comments. Could you please elaborate what you mean by "this test" and "generalize it to all linesearchers"? I already use `func testLinesearcher` to test `Bisection` and `Backtracking` so you probably have something else in mind.

vladimir-ch:
Addressed comments. I also added a check that `Linesearcher.Init` returns an evaluating operation and changed the invalid operation panic to a regular `t.Errorf`. Sorry for changing the code outside of the comments but it seemed merited to me. PTAL @btracey

functions/functions.go
```diff
@@ -1627,3 +1627,141 @@ func (Wood) Minima() []Minimum {
...
+// This function is convex but different choices of Beta1 and Beta2 lead to
+// functions with different characteristics.
```
btracey: characteristics such as?

linesearcher_test.go
```diff
@@ -0,0 +1,136 @@
...
+			return fg.Func([]float64{x})
+		},
+		g: func(x float64) float64 {
+			fg.Grad(grad, []float64{x})
+			return grad[0]
+		},
+	}
+}
+
+func testLinesearcher(ls Linesearcher, decrease, curvature float64, strongWolfe bool, t *testing.T) {
```
btracey: Could you make t first?

linesearcher_test.go
btracey: I see, I misread the test.

btracey:
Sorry, I do understand why Max and Min steps are absolute, but I don't understand why StepTolerance should be relative, especially when the other Step numbers are all absolute.

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+	bracketed bool    // Indicates if a minimum has been bracketed.
+	fInit     float64 // Function value at step = 0.
+	gInit     float64 // Derivative value at step = 0.
+
+	// When stage is 1, the algorithm updates the interval given by x and y
+	// so that it contains a minimizer of the modified function
+	//  psi(step) = f(step) - f(0) - DecreaseFactor * step * f'(0).
+	// When stage is 2, the interval is updated so that it contains a minimizer
+	// of f.
+	stage int
```
btracey: Since this only has two values, should it be a boolean?

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+
+	step         float64    // Current step.
+	lower, upper float64    // Lower and upper bounds on the next step.
```
btracey: On the next step, or just on the step?

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	if mt.MinimumStep < 0 {
+		panic("morethuente: minimum step is negative")
+	}
+	if mt.MaximumStep <= mt.MinimumStep {
+		panic("morethuente: maximum step is not greater than minimum step")
+	}
+	if mt.DecreaseFactor < 0 || mt.DecreaseFactor >= 1 {
+		panic("morethuente: invalid function tolerance")
```
btracey: invalid decrease factor would make more sense

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+	if mt.CurvatureFactor <= 0 || mt.CurvatureFactor >= 1 {
+		panic("morethuente: invalid curvature tolerance")
```
btracey: same as above

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+			// If further progress is not possible, set step to the best step
+			// obtained during the search.
+			mt.step = mt.x
+		}
+	}
+
+	return FuncEvaluation | GradEvaluation, mt.step, nil
+}
+
+// nextStep computes the next safeguarded step and updates the interval that
```
btracey: Would you like me to check this algorithm?

functions/functions.go
vladimir-ch: It's hard for me to describe it in one sentence. Please take a look at what I have there now.

morethuente.go
vladimir-ch: In fact it has three values because I (ab)use the zero value of `stage` to check that `Init` has been called. If you don't like it or if such stringent checking is deemed too pedantic, I can make it a bool.

morethuente.go
vladimir-ch: A leftover from a previous WIP code. Done.

morethuente.go
vladimir-ch: Done.

morethuente.go
vladimir-ch: I tried to factor out the interpolation code but each case differs a bit in such a way that I gave up. Still I feel that it should be possible to make it nicer, so it would be certainly nice if you could check it.

morethuente.go
vladimir-ch: I don't insist on the step tolerance being relative, it's simply what the original code is doing. What would probably make most sense would be to make it an absolute tolerance for the interval length scaled by a norm of the descent direction, something like |step_max - step_min| * ||dir|| < abs_step_tol.

morethuente.go
btracey: No, that's fine.

functions/functions.go
btracey: How about 

````
YanaiOzawaKaneko is a one-dimensional convex function where the values of Beta1 and Beta2 control the curvature around the minimum. Far away from the minimum the function approximates an absolute value function. Near the minimum, the function can either be sharply curved or flat, controlled by the parameter values"

File: morethuente.go [hunk 19]
```diff
@@ -0,0 +1,384 @@
...
+//    Decrease. ACM Transactions on Mathematical Software 20(3) (1994), 286-307
+type MoreThuente struct {
+	// DecreaseFactor is the constant factor in the sufficient decrease
+	// (Armijo) condition.
+	// It must be in the interval [0, 1). The default value is 0.
+	DecreaseFactor float64
+	// CurvatureFactor is the constant factor in the Wolfe conditions. Smaller
+	// values result in a more exact line search.
+	// A set value must be in the interval (0, 1). If it is zero, it will be
+	// defaulted to 0.9.
```
btracey: I think absolute is easier to think about. There's no clear definition of "relative" here (though I know you provided one). Additionally, the tolerance, even absolutely, is already scaled no ||dir||. Our "coordinates" have a step of size ||dir|| equal to 1.

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+		// Update x, y and step.
+		mt.nextStep(mt.fx, mt.gx, mt.fy, mt.gy, f, g)
+	}
+
+	if mt.bracketed {
+		// Monitor the length of the bracketing interval. If the interval has
+		// not been reduced sufficiently after two steps, use bisection to
+		// force its length to zero.
+		width := mt.y - mt.x
+		if math.Abs(width) >= 0.66*mt.width[1] {
```
btracey: Can we just make this 2.0/3?

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+
+// nextStep computes the next safeguarded step and updates the interval that
+// contains a step that satisfies the sufficient decrease and curvature
+// conditions.
+func (mt *MoreThuente) nextStep(fx, gx, fy, gy, f, g float64) {
+	x := mt.x
+	y := mt.y
+	step := mt.step
+
+	gSign := g
```
btracey: This is just used as a boolean (g < 0). 

gSign := g > 0
if gx < 0{
    gSign = !gSign
}

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+		// A higher function value. The minimum is bracketed between x and step.
+		// We want the next step to be closer to x because the function value
+		// there is lower.
+
+		theta := 3*(fx-f)/(step-x) + gx + g
+		s := math.Max(math.Abs(gx), math.Abs(g))
+		s = math.Max(s, math.Abs(theta))
+		gamma := s * math.Sqrt((theta/s)*(theta/s)-(gx/s)*(g/s))
+		if step < x {
+			gamma = -gamma
```
btracey: gamma *= -1?

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+		// A lower function value, derivatives of the same sign, and the
+		// magnitude of the derivative decreases. Extrapolate function values
+		// at x and step so that the next step lies between step and y.
+
+		theta := 3*(fx-f)/(step-x) + gx + g
+		s := math.Max(math.Abs(gx), math.Abs(g))
+		s = math.Max(s, math.Abs(theta))
+		gamma := s * math.Sqrt(math.Max(0, (theta/s)*(theta/s)-(gx/s)*(g/s)))
+		if step > x {
+			gamma = -gamma
```
btracey: *= -1

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+		p := (gamma - g) + theta
```
btracey: I don't think we need the parentheses.

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+		q := (gamma + (gx - g)) + gamma
```
btracey: Same here.

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+	case gSign < 0:
+		// A lower function value and derivatives of opposite sign. The minimum
+		// is bracketed between x and step. If we choose a step that is far
+		// from step, the next iteration will also likely fall in this case.
+
+		theta := 3*(fx-f)/(step-x) + gx + g
+		s := math.Max(math.Abs(gx), math.Abs(g))
+		s = math.Max(s, math.Abs(theta))
+		gamma := s * math.Sqrt((theta/s)*(theta/s)-(gx/s)*(g/s))
+		if step > x {
```
btracey: Same as below, *=-1, no parentheses

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+			// We are extrapolating so be cautious and take the step that
+			// is closer to step.
+			if math.Abs(stpc-step) < math.Abs(stpq-step) {
+				next = stpc
+			} else {
+				next = stpq
+			}
+			// Modify next if it is close to or beyond y.
+			if step > x {
+				next = math.Min(step+0.66*(y-step), next)
```
btracey: 2.0/3?

morethuente.go
```diff
@@ -0,0 +1,384 @@
...
+		// decrease rapidly in the direction of the step.
+
+		switch {
+		case mt.bracketed:
+			theta := 3*(f-fy)/(y-step) + gy + g
+			s := math.Max(math.Abs(gy), math.Abs(g))
+			s = math.Max(s, math.Abs(theta))
+			gamma := s * math.Sqrt((theta/s)*(theta/s)-(gy/s)*(g/s))
+			if step > y {
+				gamma = -gamma
```
btracey: Same as above.

functions/functions.go
vladimir-ch: Very nice description, thanks! I wanted to mention the absolute value function but couldn't formulate it in a not-too-mathematical way. Done.

morethuente.go
vladimir-ch: > Additionally, the tolerance, even absolutely, is already scaled no ||dir||. Our "coordinates" have a step of size ||dir|| equal to 1.

Yes, that's why I offered the alternative to measure the interval width in real, non-scaled-on-dir units. However, for that we would have to change the `Linesearcher` interface and that does not feel worthy. I don't have a strong argument for a relative tolerance, so I will change it to an absolute tolerance and see what it does with the tests.

morethuente.go
vladimir-ch: Done.

morethuente.go
vladimir-ch: Did s/gSign/gNeg/ and changed to bool.

morethuente.go
vladimir-ch: Done.

morethuente.go
vladimir-ch: Done.

morethuente.go
vladimir-ch: Yes, that's what I thought too. The Fortran original uses these "unnecessary" parentheses in the interpolation code consistently enough for me to think that there is a purpose (i.e., floating point magic) behind it so I kept them. I will try if removing them has any effect on the tests. At any case, I'll be happy to remove them if you want.

morethuente.go
vladimir-ch: Done.

morethuente.go
btracey: They do literally nothing though here, I’m pretty sure. The Go spec guarantees they happen in the order as written. What is says:

"Floating-point operations within a single expression are evaluated according to the associativity of the operators.  Explicit parentheses affect the evaluation by overriding the default associativity. In the expression x + (y + z) the addition y + z is performed before adding x. “

So, by default, gamma - g + theta  IS (gamma - g) + theta. I’m not sure what the Fortran rules of this are.

vladimir-ch: One fine day I should read the spec thoroughly. I will remove them.

vladimir-ch: Done.

btracey:
LGTM. Thanks.

vladimir-ch:
The discussion about the relative step tolerance has been left open-ended but I will merge anyway. I will open an issue and copy the discussion there so that it is together in one place.

PR #149, opened 2015-11-10
Use mat64 in CG

vladimir-ch:
Eventually I would like to replace floats with mat64 in all our Methods. In case of CG there is a small inconvenience because the previous gradient and direction, stored in CG as mat64.Vector, are passed to CGVariant.Beta as slices. Options:

1. Pass the slices from RawVector().Data and use floats in implementations of CGVariant (or wrap the slices again there). I don't like this much but it is what this PR does.
2. Change the CGVariant interface to take *mat64.Vector instead of slices.
3. Don't use mat64.Vector in CG.

I vote for 2. @btracey PTAL

### DIFF
--- a/cg.go
+++ b/cg.go
@@ -158,22 +158,20 @@ func (cg *CG) NextDirection(loc *Location, dir []float64) (stepSize float64) {
 
 	cg.iterFromRestart++
 	var restart bool
-	// Check if the method should be restarted because too many iterations have
-	// been taken without a restart.
 	if cg.iterFromRestart == cg.restartAfter {
-		restart = true
-	}
-	// Check if the method should be restarted because the angle between
-	// the last two gradients is too large.
-	gDot := floats.Dot(loc.Gradient, cg.gradPrev)
-	gNorm := floats.Norm(loc.Gradient, 2)
-	if gDot <= cg.AngleRestartThreshold*gNorm*cg.gradPrevNorm {
+		// Restart because too many iterations have been taken without a restart.
 		restart = true
 	}
 
-	// Compute the scaling factor β_k according to the given CG variant. Do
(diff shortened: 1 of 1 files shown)


btracey:
Removing floats where appropriate is laudable. This was clearly the case in Newton and LBFGS. I don't think removing floats everywhere is a necessary goal. We shouldn't use RawVector. mat64 documentation discourages its use for good reason (to be honest, SetRaw is much worse, but still), and we shouldn't set a bad example. 

The benefits in the other routines was clear given the interaction with matrices anway. What's the problem with floats in this code? It's just the operation of a list of numbers, and floats is better than mat64 for that.

vladimir-ch:
> This was clearly the case in Newton and LBFGS

BFGS, not LBFGS. I am looking now if LBFGS would benefit from mat64 and I think it would. The manual handling of the memory storage in LBFGS.InitDirection could be greatly simplified. NextDirection could be simpler too.

> We shouldn't use RawVector. mat64 documentation discourages its use for good reason (to be honest, SetRaw is much worse, but still), and we shouldn't set a bad example.

Agreed.

> What's the problem with floats in this code? It's just the operation of a list of numbers, and floats is better than mat64 for that.

No real problem, just using mat64.Vector is clearer and avoids calls to copy and resize.

I will drop the mat64 commit. Can I keep the other commit? The implementation comments were (and in a sense still are) unnecessarily verbose.

btracey:
LGTM

Just to be clear, I'm not opposed to mat64 changes in general, I just don't think we should be dogmatic about it in either direction. The CGVariant code seems most naturally expressed with []float64, just as optimize in general is best expressed with []float64. In other cases, keeping as a *mat64.Vector is more natural. There is some complexity in the cases that require both representations, but the overhead and complexity is small enough that I don't think we need to worry.

PR #152, opened 2015-11-10
Clean up LBFGS

vladimir-ch:
@btracey PTAL

### DIFF
--- a/lbfgs.go
+++ b/lbfgs.go
@@ -22,29 +22,32 @@ type LBFGS struct {
 
 	ls *LinesearchMethod
 
-	dim    int
-	oldest int // element of the history slices that is the oldest
-
-	x    []float64 // location at the last major iteration
-	grad []float64 // gradient at the last major iteration
-
-	a []float64 // holds cache of hessian updates
+	dim  int       // Dimension of the problem
+	x    []float64 // Location at the last major iteration
+	grad []float64 // Gradient at the last major iteration
 
 	// History
-	yHist   [][]float64 // last Store iterations of y
-	sHist   [][]float64 // last Store iterations of s
-	rhoHist []float64   // last Store iterations of rho
+	oldest int         // Index of the oldest element of the history
+	y      [][]float64 // Last Store values of y
+	s      [][]float64 // Last Store values of s
+	rho    []float64   // Last Store values of rho
+	a      []float64   // Cache of Hessian updates
 }
 
 func (l *LBFGS) Init(loc *Location) (Operation, error) {
 	if l.Linesearcher == nil {
 		l.Linesearcher = &Bisection{}
 	}
+	if l.Store == 0 {
+		l.Store = 15
+	}
+
 	if l.ls == nil {
 		l.ls = &LinesearchMethod{}
 	}
 	l.ls.Linesearcher = l.Linesearcher
 	l.ls.NextDirectioner = l
+
 	return l.ls.Init(loc)
 }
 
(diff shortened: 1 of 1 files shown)


lbfgs.go
```diff
@@ -1,52 +1,61 @@
...
 
-// LBFGS implements the limited-memory BFGS algorithm. While the normal BFGS algorithm
-// makes a full approximation to the inverse hessian, LBFGS instead approximates the
-// hessian from the last Store optimization steps. The Store parameter is a tradeoff
-// between cost of the method and accuracy of the hessian approximation.
-// LBFGS has a cost (both in memory and time) of O(Store * inputDimension).
-// Since BFGS has a cost of O(inputDimension^2), LBFGS is more appropriate
-// for very large problems. This "forgetful" nature of LBFGS may also make it perform
-// better than BFGS for functions with Hessians that vary rapidly spatially.
+// LBFGS implements the limited-memory BFGS optimization method.
```
btracey: why method instead of algorithm?

lbfgs.go
```diff
@@ -1,52 +1,61 @@
...
+//
+// While the normal BFGS algorithm makes a full approximation to the inverse
+// Hessian, LBFGS instead approximates the Hessian from the last Store
+// iterations. The Store parameter is a tradeoff between cost of the method and
```
btracey: between storage cost. Strictly it's both, but storage is the important dimension. We can leave the last line to highlight it's both.

btracey: I guess that's not true. I would delete this last sentence here and move the whole discussion into the next paragraph. This section says what it does and the next one says why

lbfgs.go
```diff
@@ -1,52 +1,61 @@
...
+// accuracy of the Hessian approximation. LBFGS has a cost (both in memory and
+// time) of O(Store * dimension).
 //
-// If Store is 0, Store is defaulted to 15.
-// A Linesearcher for LBFGS must satisfy the strong Wolfe conditions at every
-// iteration. If Linesearcher == nil, an appropriate default is chosen.
+// Since BFGS has a cost of O(dimension^2), LBFGS is more appropriate for very
```
btracey: I would flip the sentence

LBFGS is more appropriate than BFGS for large dimensional problems as the cost of LBFGS scales as O(Store * dim) while BFGS scales with O(dim^2).

lbfgs.go
```diff
@@ -1,52 +1,61 @@
...
+// large problems. The "forgetful" nature of LBFGS may also make it perform
+// better than BFGS for functions with Hessians that vary rapidly spatially.
 type LBFGS struct {
+	// Linesearcher selects suitable steps along the descent direction.
+	// Accepted steps should satisfy the strong Wolfe conditions.
+	// If Linesearcher == nil, a reasonable default will be chosen.
 	Linesearcher Linesearcher
-	Store        int // how many past iterations to store
+	// Store is the size of the limited-memory storage.
+	// If Store is 0, it will be defaulted to 15.
```
btracey: We should choose one of is and ==. @kortschak  the preferred is english in comments, right?

btracey:
LGTM

lbfgs.go
kortschak: English over symbols if possible.

lbfgs.go
vladimir-ch: Just for consistency with the other optimizers which all use the word "method".

vladimir-ch:
@btracey Could you please take one last look? I changed the LBFGS comment and I deleted the tradeoff comment. I considered deleting the word "spatially" because from the point of view of LBFGS there is no other than spatial variation but in the end I left it there.

btracey:
LGTM

PR #157, opened 2016-01-22
Rename FuncConst and GradConst in linesearchers

vladimir-ch:
Also renamed `Decrease` to `Contraction` in `Backtracking`, not sure if it is a good name. PTAL @btracey

### DIFF
--- a/backtracking.go
+++ b/backtracking.go
@@ -5,15 +5,14 @@
 package optimize
 
 const (
-	defaultBacktrackingDecrease  = 0.5
-	defaultBacktrackingFuncConst = 1e-4
-	minimumBacktrackingStepSize  = 1e-20
+	defaultBacktrackingContraction = 0.5
+	defaultBacktrackingDecrease    = 1e-4
+	minimumBacktrackingStepSize    = 1e-20
 )
 
 // Backtracking is a Linesearcher that uses backtracking to find a point that
-// satisfies the Armijo condition with the given function constant FuncConst. If
-// the Armijo condition has not been met, the step size is decreased by a
-// factor of Decrease.
+// satisfies the Armijo condition with the given decrease factor. If the Armijo
+// condition has not been met, the step size is decreased by ContractionFactor.
 //
 // The Armijo condition only requires the gradient at the beginning of each
 // major iteration (not at successive step locations), and so Backtracking may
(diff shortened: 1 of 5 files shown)


backtracking.go
```diff
@@ -5,15 +5,15 @@
...
-	defaultBacktrackingFuncConst = 1e-4
-	minimumBacktrackingStepSize  = 1e-20
+	defaultBacktrackingContraction    = 0.5
+	defaultBacktrackingDecreaseFactor = 1e-4
+	minimumBacktrackingStepSize       = 1e-20
 )
 
 // Backtracking is a Linesearcher that uses backtracking to find a point that
-// satisfies the Armijo condition with the given function constant FuncConst. If
+// satisfies the Armijo condition with the given sufficient decrease factor. If
```
btracey: rm sufficient

backtracking.go
```diff
@@ -21,12 +21,12 @@ const (
...
-// panic otherwise. If either FuncConst or Decrease are zero, it will be set to a
-// reasonable default.
+// Both DecreaseFactor and Contraction must be between zero and one, and
+// Backtracking will panic otherwise. If either DecreaseFactor or Contraction
+// are zero, it will be set to a reasonable default.
 type Backtracking struct {
-	FuncConst float64 // Necessary function descrease for Armijo condition.
-	Decrease  float64 // Step size multiplier at each iteration (stepSize *= Decrease).
+	DecreaseFactor float64 // Constant factor in the sufficient decrease (Armijo) condition.
+	Contraction    float64 // Step size multiplier at each iteration (step *= Contraction).
```
btracey: ContractionFactor?

backtracking.go
```diff
@@ -5,15 +5,15 @@
...
+	defaultBacktrackingDecreaseFactor = 1e-4
+	minimumBacktrackingStepSize       = 1e-20
 )
 
 // Backtracking is a Linesearcher that uses backtracking to find a point that
-// satisfies the Armijo condition with the given function constant FuncConst. If
+// satisfies the Armijo condition with the given sufficient decrease factor. If
 // the Armijo condition has not been met, the step size is decreased by a
-// factor of Decrease.
+// factor of Contraction.
```
btracey: is decreased by ContractionFactor?

bisection.go
```diff
@@ -7,11 +7,14 @@ package optimize
 import "math"
 
 // Bisection is a Linesearcher that uses a bisection to find a point that
-// satisfies the strong Wolfe conditions with the given gradient constant and
-// function constant of zero. If GradConst is zero, it will be set to a reasonable
-// value. Bisection will panic if GradConst is not between zero and one.
+// satisfies the strong Wolfe conditions with the given curvature factor and
+// sufficient decrease factor of zero.
```
btracey: curvature factor and no function decrease.

btracey:
LGTM with minor comments.

backtracking.go
vladimir-ch: Done.

backtracking.go
vladimir-ch: Done.

backtracking.go
vladimir-ch: Done.

bisection.go
vladimir-ch: I think that saying "no function decrease" without "factor" could be misleading/incorrect because in fact there is no function _increase_. Finishing the sentence after "Wolfe conditions" should be enough clear, we talk about the factors because we want to stress that there is only the curvature factor. But even if we use the decrease factor of zero, there is no reason to prevent the users from setting it to a positive value. So unless you object, in another PR I'd like to add DecreaseFactor to Bisection and trim the sentence after "Wolfe conditions". I'll wait for your comment before merging this PR, @btracey

btracey: Yea, there should probably be a DecreaseFactor. The internal logic may need to change some though, so if you want to do that in a different PR that's fine.

Issue #155, opened 2016-01-22
Change FunConst and GradConst to DecreaseFactor and CurvatureFactor


btracey:
Much nicer names. Already in MoreThuente, should be changed in other linesearchers.

PR #158, opened 2016-01-26
Small clean-up in Bisection

vladimir-ch:
I looked into adding `DecreaseFactor` to `Bisection` as discussed in #155. I am not sure if the logic of iterating towards the minimum would work under all conditions with a positive `DecreaseFactor`. I think that it would consists of more than just modifying an `if` check and passing a non-zero to `StrongWolfeConditionsMet`. Quite possibly a two-stage approach like in `MoreThuente` would be needed. Therefore I abandon the plan for the moment.

So at least I did a tiny cleanup. PTAL @btracey

### DIFF
--- a/bisection.go
+++ b/bisection.go
@@ -6,9 +6,13 @@ package optimize
 
 import "math"
 
+const (
+	defaultBisectionCurvature = 0.9
+)
+
 // Bisection is a Linesearcher that uses a bisection to find a point that
 // satisfies the strong Wolfe conditions with the given curvature factor and
-// sufficient decrease factor of zero.
+// a decrease factor of zero.
 type Bisection struct {
 	// CurvatureFactor is the constant factor in the curvature condition.
 	// Smaller values result in a more exact line search.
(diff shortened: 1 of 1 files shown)


bisection.go
```diff
@@ -75,7 +79,7 @@ func (b *Bisection) Iterate(f, g float64) (Operation, float64, error) {
 		// See if the function value is good enough to make progress. If it is,
 		// evaluate the gradient. If not, set it to the upper bound if the bound
 		// has not yet been found, otherwise iterate toward the minimum location.
-		if f <= b.minF {
+		if f <= minF {
```
vladimir-ch: I think that this is the correct condition but please check.

bisection.go
```diff
@@ -6,9 +6,13 @@ package optimize
...
 import "math"
 
+const (
+	defaultBisectionCurvature = 0.9
+)
+
 // Bisection is a Linesearcher that uses a bisection to find a point that
 // satisfies the strong Wolfe conditions with the given curvature factor and
-// sufficient decrease factor of zero.
+// the decrease factor of zero.
```
btracey: a decrease factor?

btracey:
LGTM

PR #163, opened 2016-04-18
go 1.5.3 -> 1.5.4, 1.6 -> 1.6.1

jonlawlor:


btracey:
LGTM

jonlawlor:
It appears that there is an issue with:
```
TestBrownBadlyScaled
TestExtendedRosenbrock
TestPowelBadlyScaled
TestMinimalSurface
```

I don't know enough about this repo at the moment to diagnose, but I think the most likely cause is a change to a gonum library that it depends on. (matrix or floats?)

btracey:
Yea, it's probably because the problems are badly scaled :). @vladimir-ch

jonlawlor:
I think the change is working as intended, and a different PR should be opened to deal with the tests.  Can you re-LGTM, and then I will open an issue?

vladimir-ch:
Sorry, I missed this thread. The failures will be caused by the changed default step size in fd.Gradient, not by this PR, so LGTM. I'll have a look at the failures tomorrow.

jonlawlor:
No worries - the changes to go version should not have any impact on any gonum project as far as I know, so this is just housekeeping.

PR #165, opened 2016-04-21
Add package for linear programming solving and Simplex implementation

btracey:


btracey:
PTAL @vladimir-ch 

There are lots and lots of minor cleanups that can be done. I'm afraid that if too many are done, it'll be hard to understand the algorithm at the first pass. I figure like this you have a hope of doing a look-through and afterwards cleanup can happen. 

It's been a while since I was working with this code, but at the time I had done extensive testing on it. There are still cases that fail, though some of them also cause the matlab LP solver to crash. You can see that the tests are generating a 5 digit number of tests, so I'm pretty sure the algorithm is correct (if not perfectly robust to numeric error).

convex/lp/convert.go
```diff
@@ -0,0 +1,137 @@
+package lp
```
sbinet: license header?

convex/lp/simplex.go
```diff
@@ -0,0 +1,601 @@
+package lp
```
sbinet: license header?

convex/lp/simplex_test.go
```diff
@@ -0,0 +1,268 @@
+package lp
```
sbinet: license header?

sbinet:
I would perhaps add an example:

```go
// Copyright ©2016 The gonum Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package lp_test

import (
	"fmt"
	"log"

	"github.com/gonum/matrix/mat64"
	"github.com/gonum/optimize/convex/lp"
)

func Example_Simplex() {
	c := []float64{-1, -2, 0, 0}
	A := mat64.NewDense(2, 4, []float64{-1, 2, 1, 0, 3, 1, 0, 1})
	b := []float64{4, 9}

	opt, x, err := lp.Simplex(c, A, b, 0, nil)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("opt: %v\n", opt)
	fmt.Printf("x: %v\n", x)
	// Output:
	// opt: -8
	// x: [2 3 0 0]
}
```

tested with `scipy.optimize.linprog`:
http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html#scipy.optimize.linprog

convex/lp/simplex.go
```diff
@@ -0,0 +1,601 @@
...
+// must equal the number of rows of A. Simplex will panic if these conditions are
+// not met.
+//
+// initialBasic can be used to set the initial set of indices for a feasible
+// solution to the LP. If an initial feasible solution is not known, initialBasic
+// may be nil. If initialBasic is non-nil, len(initialBasic) must equal the number
+// of columns of A and must be an actual feasible solution to the LP, otherwise
+// Simplex will panic.
+//
+// For a detailed introduction to the Simplex method please see lectures 11-13 of
```
sbinet: perhaps also add an article/book reference?

btracey: Do you have one in mind? In the documentation, I like to link references that are actually useful, and I learned the simplex from those lectures.

sbinet: the canonical one seems to be:
```
J.A. Nelder and R. Mead. A simplex method for function minimization. Comput. J., 7:308, 1965.
```

in my field, we usually just refer to `MINUIT`:
- https://en.wikipedia.org/wiki/MINUIT
- http://seal.web.cern.ch/seal/documents/minuit/mntutorial.pdf

convex/lp/convert.go
btracey: Done.

convex/lp/simplex.go
btracey: Done.

convex/lp/simplex.go
btracey: Wrong simplex method :). This is Danzig's Simplex, not [Nelder-Mead](https://godoc.org/github.com/gonum/optimize#NelderMead)

btracey: Done.

convex/lp/simplex_test.go
btracey: Done.

convex/lp/simplex.go
sbinet: ha! :)

kortschak: In the opening sentence of the doc comment it looks like you are aligning with tabs. Please do that with spaces; tabs are not always rendered the same width on all platforms.

convex/lp/simplex_test.go
```diff
@@ -0,0 +1,272 @@
...
+		if primalBad {
+			t.Errorf("non-known error returned: %s", errPrimal)
+		}
+		return errPrimal
+	}
+
+	// Compare the result to the answer found from solving the dual LP.
+
+	// Construct and solve the dual LP.
+	// Standard Frm:
```
sbinet:
```
-Standard Frm
+Standard Form
```

convex/lp/simplex.go
```diff
@@ -0,0 +1,608 @@
...
+			// Either unbounded or something went wrong.
+			return -1, -1, err
+		}
+		replace = floats.MinIdx(move)
+		if math.Abs(move[replace]) > blandZeroTol {
+			// Large enough that it shouldn't be a problem
+			return replace, minIdx, nil
+		}
+		// Find a zero index where replacement is non-singular.
+		biCopy := make([]int, len(basicIdxs))
```
vladimir-ch: This allocation can be moved out from the loop.

convex/lp/simplex.go
```diff
@@ -0,0 +1,608 @@
...
+		}
+	}
+	if !allPos {
+		return xb, errors.New("lp: supplied subcolumns not a feasible solution")
+	}
+	return xb, nil
+}
+
+// extractColumns creates a new matrix out of the columns of A specified by cols.
+func extractColumns(A mat64.Matrix, cols []int) *mat64.Dense {
```
vladimir-ch: It seems to me that this could take the result as a parameter and thus avoid creating garbage at a few places.

convex/lp/simplex.go
```diff
@@ -0,0 +1,608 @@
...
+	// the intersection of several constraints. Use the Bland rule instead
+	// of the rule in step 4 to avoid cycling.
+	for {
+		// Compute reduced costs -- r = cn - an^T ab^-T cb
+		var tmp mat64.Vector
+		err = tmp.SolveVec(ab.T(), cbVec)
+		if err != nil {
+			break
+		}
+		data := make([]float64, n-m)
```
vladimir-ch: This can be moved out of the loop.

convex/lp/simplex.go
```diff
@@ -0,0 +1,608 @@
...
+
+	// Temporary data needed each iteration. (Described later)
+	r := make([]float64, n-m)
+	move := make([]float64, m)
+
+	// Solve the linear program starting from the initial feasible set. This is
+	// the "Phase 2" problem.
+	//
+	// Algorithm:
+	// 1) Compute the "reduced costs" for the non-basic variables. The reduced
```
vladimir-ch: I would prefer to have these comments distributed to corresponding places inside the loop body. The body is not so big and having the descriptions closer to the actual code is clearer, at least for me. Up to you.

convex/lp/simplex.go
```diff
@@ -0,0 +1,608 @@
...
+		basicIdxs[replace], nonBasicIdx[minIdx] = nonBasicIdx[minIdx], basicIdxs[replace]
+		cb[replace], cn[minIdx] = cn[minIdx], cb[replace]
+		tmpCol1 := mat64.Col(nil, replace, ab)
+		tmpCol2 := mat64.Col(nil, minIdx, an)
+		ab.SetCol(replace, tmpCol2)
+		an.SetCol(minIdx, tmpCol1)
+
+		// Compute the new xb.
+		xbVec := mat64.NewVector(len(xb), xb)
+		err = xbVec.SolveVec(ab, bVec)
```
vladimir-ch: It seems to me that the matrix factorized here and in the next iteration inside computeMove is the same, so one factorization can be avoided. I'm not sure how much work that would be, a future PR is ok.

vladimir-ch:
I quickly skimmed through the code and pointed out a couple of things that I saw. I think that the algorithm is quite clear from the code although I didn't go into the details. The algorithm can be surely made more efficient by updating the factorization but given our limited manpower that's a future work. There are many possible cleanups, especially in the docs, too many to point out individually. So I think it is best to merge this and fix things in another PR. So LGTM.

convex/lp/simplex.go
btracey: The code does have simpler versions of the comment within the for loop , i.e. `Compute reduced costs -- r = cn - an^T ab^-T cb`.

convex/lp/simplex.go
btracey: Added TODO.

convex/lp/simplex_test.go
btracey: Done.

btracey:
I added your example @sbinet

PR #167, opened 2016-05-10
optimize: add example for Local+BFGS

sbinet:
This CL adds a simple example, extracted from:
 http://docs.scipy.org/doc/scipy-0.17.0/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize

### DIFF
--- a/local_example_test.go
+++ b/local_example_test.go
@@ -1,0 +1,43 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package optimize_test
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gonum/optimize"
+	"github.com/gonum/optimize/functions"
+)
+
+func ExampleLocal_BFGS() {
+	p := optimize.Problem{
+		Func: functions.ExtendedRosenbrock{}.Func,
+		Grad: functions.ExtendedRosenbrock{}.Grad,
+	}
+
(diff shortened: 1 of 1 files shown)


local_example_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/gonum/optimize/functions"
+)
+
+func ExampleLocal_BFGS() {
+	p := optimize.Problem{
+		Func: functions.ExtendedRosenbrock{}.Func,
+		Grad: functions.ExtendedRosenbrock{}.Grad,
+	}
+
+	method := &optimize.BFGS{}
```
btracey: I'd move this either in the call to local or just before the call to local, that way it's a bit more obvious.

local_example_test.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	if err = result.Status.Err(); err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("result.Status: %v\n", result.Status)
+	fmt.Printf("result.X: %v\n", result.X)
+	fmt.Printf("result.F: %v\n", result.F)
+	// Output:
```
btracey: Maybe add the number of function evaluations? That's going to be the relevant quantity between methods.

btracey:
LGTM

local_example_test.go
sbinet: done

local_example_test.go
sbinet: done

PR #172, opened 2016-06-09
Add Predictor-Corrector and Self-Dual Interior Point solvers

dane-unltd:
The interior point solvers I have written some time ago.
There is a lot of code in there and not much documentation ...
It is mostly based on the descriptions in
http://www.siam.org/books/swright/
with some additional information on the self-dual approach from papers like
http://link.springer.com/article/10.1007%2FBF02206815

The self-dual algorithm is preferred since it deals better with infeasible problems.
I do not actually use the code, but both methods seem to work satisfactory. However, using the same tests as for the simplex algorithm, I had to reduce the tolerances when checking the output. The tolerances should probably be fixed anyway and not chosen by the user.

If you guys think that this would be a useful addition to the optimization package I will leave this pull-request open and try to clean up the code and add some documentation when I got some time.


convex/lp/ip_simple_test.go
```diff
@@ -0,0 +1,186 @@
...
+			resP.MulVec(A, x)
+			resP.SubVec(b, resP)
+
+			// resS = - x(*)s
+			resS.MulElemVec(x, s)
+			resS.ScaleVec(-1, resS)
+
+			dev = (asum(resD) + asum(resP) + asum(resS)) / float64(n)
+		}
+		if (dev >= 500*tol) && err == nil {
```
sbinet: should probably be `err != nil`

vladimir-ch:

I think it would be worth having this class of algorithms in Gonum, so @dane-unltd please feel free to submit a new PR in the new repository if you are still around and interested.

PR #173, opened 2016-06-10
Simplify findLinearlyIndependent

btracey:


btracey:
PTAL @vladimir-ch @sbinet

convex/lp/simplex.go
```diff
@@ -578,32 +578,19 @@ func findLinearlyIndependent(A mat64.Matrix) []int {
 			break
 		}
 		mat64.Col(newCol, i, A)
+		columns.SetCol(len(idxs), newCol)
 		if len(idxs) == 0 {
 			// A column is linearly independent from the null set.
-			// This is what needs to be changed if zero columns are allowed, as
-			// a column of all zeros is not linearly independent from itself.
-			columns.SetCol(len(idxs), newCol)
+			// This needs to change if zero columns are allowed, as
```
sbinet: this comment is a bit confusing (to me at least).
it reads as if this piece of code needs to be changed, ie: as a TODO/FIXME.

perhaps:
```go
// If zero columns are allowed and as a column of all zeros is 
// not linearly independent with itself,
// mark this column as needing to be changed.
```

sbinet:
LGTM

PR #175, opened 2016-06-10
Refactor extractColumns to have a destination matrix

btracey:


btracey:
PTAL @sbinet @vladimir-ch

convex/lp/simplex.go
```diff
@@ -349,7 +352,8 @@ func replaceBland(A mat64.Matrix, ab *mat64.Dense, xb []float64, basicIdxs, nonB
 			}
 			copy(biCopy, basicIdxs)
 			biCopy[replace] = nonBasicIdx[minIdx]
-			abTmp := extractColumns(A, biCopy)
+			abTmp := mat64.NewDense(m, len(biCopy), nil)
```
btracey: I know this can be refactored out of the loop, but I'd rather do it as an optimization pass on replaceBland and keep this PR about extractColumns

convex/lp/simplex.go
```diff
@@ -443,17 +447,21 @@ func initializeFromBasic(ab *mat64.Dense, b []float64) (xb []float64, err error)
...
-// TODO(btracey): Allow this to take a receiver.
-func extractColumns(A mat64.Matrix, cols []int) *mat64.Dense {
-	r, _ := A.Dims()
-	sub := mat64.NewDense(r, len(cols), nil)
+// extractColumns copies the columns specified by cols into the columns of dst.
+func extractColumns(dst *mat64.Dense, A mat64.Matrix, cols []int) {
+	r, c := dst.Dims()
+	ra, _ := A.Dims()
+	if ra != r {
+		panic("simplex: row mismatch")
```
sbinet: shouldn't this be `"lp: row mismatch"` instead?
we don't seem to be too consistent in `gonum/optimize/...`, but perhaps we could start :)

sbinet:
LGTM

convex/lp/simplex.go
btracey: My thought was that this panic should never be triggered -- it can only be accessed internally. Thus, the panic purely belongs to simplex.

sbinet: fair enough.

PR #176, opened 2016-06-11
Cleanup findInitialBasic and initializeFromBasic

btracey:


btracey:
PTAL @sbinet

convex/lp/simplex.go
```diff
@@ -528,44 +534,53 @@ func findInitialBasic(A mat64.Matrix, b []float64) ([]int, *mat64.Dense, []float
...
+	var set bool
 	for i := range xOpt {
 		if _, inBasic := basicMap[i]; inBasic {
 			continue
 		}
-		newBasic[basicIdx] = i
-		extractColumns(ab, A, newBasic)
-		xb, err := initializeFromBasic(ab, b)
+		newBasic[addedIdx] = i
+		if set {
```
sbinet: this is probably just very nit-picking but I would probably reverse the if-else-branch:

```go
if !set {
   set = true
   extractColumns(ab, A, newBasic)
} else {
   mat64.Col(col, i, A)
   ab.SetCol(addedIdx, col)
}
```

that way, the test on `set` and its first (and last) modification are closer, visually.

(when I read it first, I was a bit confused whether this wasn't just dead code...)

sbinet:
just a little nitpick, but LGTM otherwise.

convex/lp/simplex.go
btracey: Done.

btracey: Actually, I accidentally nuked my local version of this... I'll submit this in the next PR

PR #183, opened 2016-08-11
Add Global optimization method and GuessAndCheck algorithm

btracey:


btracey:
@vladimir-ch  Here is another vision of the global optimization routine. Like last time, it's in a 30% done stage. Rather than detailed comments, looking at opinions on high-level structure (sufficiently clean, think it'll expand, etc.)

global.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+	}
+}
+
+// globalOperation updates handles the status received by an individual worker.
+// It uses a mutex to protect updates where necessary.
+func (g *globalStatus) globalOperation(op Operation, loc *Location, x []float64) Status {
+	// TODO(btracey): Combine this with Local
+	stats := g.stats
+	mux := g.mux
```
sbinet: (I know you only want high-level comments, but bear with me...)

copying `sync.Mutex` is bound to raise eyebrows.
I know here we just copy a `*sync.Mutex` but... couldn't we remove this line and just access the mutex _via_ `g.mux` ?
I'd feel much less woried about a possible modification of `globalStatus` down the line...

btracey: Agreed. The original code had mux as a function input, and so this was the easiest change to get something to compile. This way seemed slightly easier to get the flavor of what's going on, so I left it. This will definitely be changed.

global.go
```diff
@@ -0,0 +1,248 @@
...
+	"time"
+
+	"github.com/gonum/matrix/mat64"
+)
+
+// GlobalMethod is a global optimizer. Typically will require more function
+// evaluations and no sense of local convergence
+type GlobalMethod interface {
+	// Need to tell the optimizer how many can be done at once (for example random restart)
+	InitGlobal(tasks int) // can this be concurrent bool? Seems like optimizer should be told how many tasks
```
vladimir-ch: I think that passing an int here is ok, perhaps `Settings.Concurrent` could be a bool.

global.go
```diff
@@ -0,0 +1,248 @@
...
+// Global uses a global optimizer to search for the gloabl minimum of a
+// function. A maximization problem can be transformed into a
+// minimization problem by multiplying the function by -1.
+//
+// Global, unlike Local, searches for the global minimum of the function.
+// GlobalMethods typically do not make any assumptions about the number and location
+// of local minima, and so the only convergence metric used is the convergence
+// of function values found at major iterations. GradientThreshold is not checked
+// during convergence, and initial values are not used.
+func Global(p Problem, dim int, settings *Settings, method GlobalMethod) (*Result, error) {
```
vladimir-ch: It is natural but still nice that it is not possible to pass any initial guess here.

global.go
```diff
@@ -0,0 +1,248 @@
...
+	if method == nil {
+		// TODO(btracey): Replace with a better global minimizer (basically
+		// anything else). Though technically this isn't worse on the space of
+		// all problems.
+		method = &GuessAndCheck{}
+	}
+	if settings == nil {
+		settings = DefaultSettingsGlobal()
```
vladimir-ch: I think we could give meaning to the zero value of `Settings` like we did in `diff`.

global.go
```diff
@@ -0,0 +1,248 @@
...
+			defer wg.Done()
+			loc := newLocation(dim, method)
+			x := make([]float64, dim)
+			globalWorker(task, method, gs, loc, x)
+		}(task)
+	}
+	wg.Wait()
+
+	// Cleanup and collect results
+	if settings.Recorder != nil && err == nil {
```
vladimir-ch: Low-level remark: `err` is always nil here because it has not been modified after the call to `initializeOptimization`.

global.go
```diff
@@ -0,0 +1,248 @@
...
+	optLoc    *Location
+	settings  *Settings
+}
+
+func globalWorker(task int, m GlobalMethod, g *globalStatus, loc *Location, x []float64) {
+	for {
+		// Find Evaluation location
+		op, err := m.IterateGlobal(task, loc)
+		if err != nil {
+			// TODO(btracey): Figure out how to handle errors properly. Shut
```
vladimir-ch: How about something similar to https://godoc.org/golang.org/x/sync/errgroup?

guessandcheck.go
```diff
@@ -0,0 +1,33 @@
...
+
+func (g *GuessAndCheck) Needs() struct{ Gradient, Hessian bool } {
+	return struct{ Gradient, Hessian bool }{false, false}
+}
+
+func (g *GuessAndCheck) InitGlobal(tasks int) {
+	g.eval = make([]bool, tasks)
+}
+
+func (g *GuessAndCheck) IterateGlobal(task int, loc *Location) (Operation, error) {
```
vladimir-ch: I like how this has simplified compared to #146

vladimir-ch:
On high level, I like what is in here, I can't think of anything I would change in the important parts of the API, `Global` and `GlobalMethod`.

On low implementation level, I think we should and can avoid the synchronization via mutex, and we need to figure out the error handling.

global.go
btracey: Yea. Agreed.

global.go
btracey: The challenge is the minimum function val.

global.go
btracey: Whoa. Did not know that was a thing.

Barring that, I think it should probably shut down the process. The method is in charge of returning the error, so if it wants to handle an error internally it can.

guessandcheck.go
btracey: Yea.

btracey:
We should reduce the mutex overlap as much as possible, but the fundamental problem is unavoidable. The main thing is to make sure function evaluations aren't blocking other function evaluations, which is the case here.

global.go
btracey: We could make it that you have to use 1e-16 (or whatever) to signify zero, unclear which is worse

btracey:
PTAL @vladimir-ch 

This is a full implementation, I think. There is obviously lots of parallel between this and Local, which is good. However, the two are subtly different in ways that make it difficult to simply combine the two.  What I would like to do in follow-on PRs is:
0) Get this change (or something like it submitted)
1) Significantly combines the implementations of Local and Global
2) Add better tests for Global, and improve comments where needed.
3) Add a "real" global optimizer

The above a) Isolates adding Global from messing up local. b) Reduces unnecessary comment work. If you disagree with the spirit of b, I can make a bunch of comment improvements in this PR.

One open question is if OptLoc should always be updated on major iteration, like it does in Local. I lean toward yes for consistency (changing both here and Global). 

I think I disagree with avoiding a mutex. We have to have serialization to update stats, etc. This has to be done with a mutex or channel. It's not clear that a channel is at all better. This version at least has no blocking during function evaluation which is the main thing.

errors.go
```diff
@@ -28,6 +28,8 @@ var (
 	// due to floating-point arithmetic.
 	ErrNoProgress = errors.New("linesearch: no change in location after Linesearcher step")
 
+	ErrFunctionConvergeNil = errors.New("global: function convergence must not be nil")
```
vladimir-ch: This error is not used anywhere.

global.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+	}
+
+	stats := &Stats{}
+	optLoc := newLocation(dim, method)
+	optLoc.F = math.Inf(1)
+
+	if settings.FunctionConverge != nil {
+		settings.FunctionConverge.Init(optLoc.F)
+	}
```
vladimir-ch: The code needs FunctionConverge to be non-nil, panic here?

global.go
```diff
@@ -0,0 +1,328 @@
...
+
+// GlobalMethod is a global optimizer. Typically will require more function
+// evaluations and no sense of local convergence
+type GlobalMethod interface {
+	// Global tells method the max number of tasks, method returns how many it wants.
+	// This is needed to sync the Global goroutines and inside goroutines.
+	InitGlobal(tasks int) int
+	// Global method may assume that the same task id always has the same pointer with it.
+	IterateGlobal(task int, loc *Location) (Operation, error)
+	Needs() struct{ Gradient, Hessian bool }
```
vladimir-ch: Needser

errors.go
vladimir-ch: Maybe it should be just a panic string.

global.go
```diff
@@ -0,0 +1,328 @@
...
+		// time-consuming) evaluation at the starting location.
+		err := settings.Recorder.Init()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	stats := &Stats{}
+	optLoc := newLocation(dim, method)
+	optLoc.F = math.Inf(1)
```
vladimir-ch: Make this part of `newLocation`?

global.go
```diff
@@ -0,0 +1,328 @@
...
+	if settings.FunctionConverge != nil {
+		settings.FunctionConverge.Init(optLoc.F)
+	}
+
+	stats.Runtime = time.Since(startTime)
+
+	// Don't need to check convergence because it can't possibly have converged.
+	// (No function evaluations and no starting location).
+	var err error
+	var status Status
```
vladimir-ch: Move this before the call to `minimizeGlobal`

vladimir-ch:
I've given it a quick pass and LGTM with minor comments, I think it turned out quite nicely.

I've ignored the comments, please improve them later.

> One open question is if OptLoc should always be updated on major iteration, like it does in Local. I lean toward yes for consistency (changing both here and Global).

Yes, I think it should be on major iteration like it is in Local. There we agreed that it is the optimizer's responsibility to tell us when the location is complete and ready for a convergence check. Here I think the situation is the same.

> I think I disagree with avoiding a mutex. We have to have serialization to update stats, etc. ...

It is ok to have the mutex.

global.go
btracey: Originally I thought so, but now I don't think so. It's pretty normal to say "Run for 500 iterations". The question is if we should test to make sure it can eventually converge, or leave that up to the user. I think we should not test, because there are a bunch of ways (method error, statuser, etc)

vladimir-ch: Ok, but then it will panic in `checkConvergenceGlobal`, so perhaps there should be a check for non-nil there.

btracey: Yea, agreed.

errors.go
btracey: Deleted.

global.go
btracey: Done.

global.go
btracey: Done.

global.go
btracey: Done.

global.go
btracey: Done.

PR #196, opened 2017-02-09
Add FullRecorder for logging an entire optimization

btracey:


btracey:
I'm writing some personal code now to verify.

I'm not sure "Read" is the function name. Maybe Read/RealAll as a parallel to csv?

btracey:
If we do change to Read/ReadAll, do we change the struct to also have an io.Reader? Do we make a different type entirely?

btracey:
This doesn't record the initial optimization step.

recorder.go
```diff
@@ -104,3 +105,72 @@ func (p *Printer) Record(loc *Location, op Operation, stats *Stats) error {
 	p.lastValue = time.Now()
 	return nil
 }
+
+// Full Recorder records all of the evaluations that occur during an optimization
```
vladimir-ch: Do you have a justification why this should be in optimize? The functionality seems quite specific to fit general needs.

btracey: It seems specific? In what way? It seems like a natural counterpart to Printer. Printer prints to a writer to track optimization, this saves the optimization run to `json` for retrospection

vladimir-ch: I was thinking about what is being recorded and into what format. I can imagine that either will easily not fit the needs and the user will write it's own recorder.

recorder.go
```diff
@@ -104,3 +105,72 @@ func (p *Printer) Record(loc *Location, op Operation, stats *Stats) error {
...
+
+func (f FullRecorder) Record(loc *Location, op Operation, stats *Stats) error {
+	r := RecordLocation{Op: op.String()}
+	switch {
+	case op.isEvaluation():
+		r.Loc.X = loc.X
+		switch op {
+		default:
+			panic("optimize: unknown evaluation operation")
+		case FuncEvaluation:
```
vladimir-ch: Evaluation operations can be ORed together which this switch does not take into account.

recorder.go
```diff
@@ -104,3 +105,72 @@ func (p *Printer) Record(loc *Location, op Operation, stats *Stats) error {
...
+			r.Loc.F = loc.F
+		case GradEvaluation:
+			r.Loc.Gradient = loc.Gradient
+		case HessEvaluation:
+			r.Loc.Hessian = loc.Hessian
+		}
+	case op == MajorIteration || op == PostIteration:
+		r.Loc = *loc
+	}
+	if op.isEvaluation() {
```
vladimir-ch: A left-over before the above switch was introduced?

recorder.go
```diff
@@ -104,3 +105,72 @@ func (p *Printer) Record(loc *Location, op Operation, stats *Stats) error {
...
+	}
+	b, err := json.MarshalIndent(r, "", "\t")
+	if err != nil {
+		return nil
+	}
+	_, err = f.Writer.Write(b)
+	return err
+}
+
+// Read reads the stream written to by FullRecorder, and returns the history of
```
vladimir-ch: `written to r` ?

recorder.go
vladimir-ch: `FullRecorder`

recorder.go
```diff
@@ -104,3 +105,72 @@ func (p *Printer) Record(loc *Location, op Operation, stats *Stats) error {
 	p.lastValue = time.Now()
 	return nil
 }
+
+// Full Recorder records all of the evaluations that occur during an optimization
+// run. If Operation is an Evaulation, FullRecorder records the x location and
+// the corresponding field of Location. If Operation is a MajorIteration, or
```
vladimir-ch: field or fields

recorder.go
btracey: I understand your point about things being Ored and this not handling that properly. Do you have other suggestions for making it more general? It seems reasonable to me to provide some built-in way to store the history of the optimization for retrospection (in addition to the human-friendly printing we have now).

btracey:
Okay, I think I see your point about Json being speciifc. Maybe the full recorder can just keep a list of the steps? That was the user can do something with them in memory, or save them using Json after the optimization has concluded (or whatever else).

vladimir-ch:

